PR #3, opened 2022-01-12
Solution

Abnormaltype:


app/main.py
```diff
@@ -1,13 +1,30 @@
...
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # white your code here
-    pass
+    return {"errors": [
+        {"line": errors[i]["line_number"],
+         "column": errors[i]["column_number"],
+         "message": errors[i]["text"],
+         "name": errors[i]["code"],
+         "source": "flake8"}
```
MasterpieceElbow, line 15: You have implemented "format_linter_error" above, why did you duplicate code from that function?

app/main.py
```diff
@@ -1,13 +1,30 @@
...
-    pass
+    return [{"errors": [
+        {"line": v[i]["line_number"],
+         "column": v[i]["column_number"],
+         "message": v[i]["text"],
+         "name": v[i]["code"],
+         "source": "flake8"}
+        for i in range(len(v))], "path": k, "status": "failed"} if len(v) != 0
+        else {"errors": v, "path": k, "status": "passed"}
+        for k, v in linter_report.items()]
```
MasterpieceElbow, line 30: Don't repeat yourself, use your implemented functions!

MasterpieceElbow (changes requested):

app/main.py
Abnormaltype: okay. but I thought the point was to realize them from scratch

app/main.py
Abnormaltype: okay, I know rule DRY thx) I've fixed

app/main.py
MasterpieceElbow: No, main concept of function is to encapsulate code inside and reuse it

MasterpieceElbow (approved):

app/main.py
Abnormaltype: thx, sorry, it's my mistake

PR #20, opened 2022-01-14
'Solution'

serhii56465:


app/main.py
```diff
@@ -1,13 +1,25 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # white your code here
-    pass
+    return {"errors": [{"line": error["line_number"],
+                        "column": error["column_number"],
+                        "message": error["text"],
+                        "name": error["code"], "source": "flake8"}
+                       for error in errors],
```
MasterpieceElbow, line 12: You have already wrote function that format single error, why do you repeat yourself?) Use that function instead.

app/main.py
```diff
@@ -1,13 +1,25 @@
...
 def format_linter_report(linter_report: dict) -> list:
-    # white your code here
-    pass
+    return [{"errors": [{"line": error["line_number"],
+                         "column": error["column_number"],
+                         "message": error["text"],
+                         "name": error["code"], "source": "flake8"}
+                        for error in errors],
+             "path": file_path,
+             "status": "passed" if len(errors) == 0 else "failed"}
```
MasterpieceElbow, line 24: The same as previous comment, don't repeat yourself! Almost all work you do in 'format_single_linter_file'.

PR #21, opened 2022-01-15
Solution

Alexandr-Semenishin:


app/main.py
```diff
@@ -1,13 +1,10 @@
 def format_linter_error(error: dict) -> dict:
-    # white your code here
-    pass
+    return {"line": error["line_number"], "column": error["column_number"], "message": error["text"], "name": error["code"], "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # white your code here
-    pass
+    return {"errors": [{"line": er["line_number"], "column": er["column_number"], "message": er["text"], "name": er["code"], "source": "flake8"} for er in errors if er["filename"] == file_path], "path": file_path, "status": "failed"}
```
MasterpieceElbow, line 6: Why do you not use you function for formatting single error? You duplicate your code. Remake.

app/main.py
```diff
@@ -1,13 +1,10 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # white your code here
-    pass
+    return [{"errors": [{"line": er["line_number"], "column": er["column_number"], "message": er["text"], "name": er["code"], "source": "flake8"} for er in value], "path": key, "status": "passed" if value == [] else "failed"} for key, value in linter_report.items()]
```
MasterpieceElbow, line 10: The same as 'format_single_linter_file'

app/main.py
```diff
@@ -1,13 +1,10 @@
 def format_linter_error(error: dict) -> dict:
-    # white your code here
-    pass
+    return {"line": error["line_number"], "column": error["column_number"], "message": error["text"], "name": error["code"], "source": "flake8"}
```
MasterpieceElbow, line 2: For all functions: beautify your return, make it look like:
return {
               key: value,
               key: value,
               key: value
           }

MasterpieceElbow (changes requested):

PR #39, opened 2022-05-11
'Solution'

Aleksei-Isaev:


app/main.py
```diff
@@ -1,13 +1,16 @@
...
+            'column': error['column_number'],
+            'message': error['text'],
+            'name': error['code'],
+            'source': 'flake8'}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # white your code here
-    pass
+    return {'errors': [format_linter_error(name) for name in errors],
```
MasterpieceElbow, line 10: `name` is not a good variable name for dict with error information.

app/main.py
```diff
@@ -1,13 +1,16 @@
...
+            'path': file_path,
+            'status': 'failed' if errors else 'passed'}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # white your code here
-    pass
+    return [format_single_linter_file(i, j) for i, j in linter_report.items()]
```
MasterpieceElbow, line 16: `i` and `j` are bad names. Name them according to what they mean in `linter_report.items()`

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #40, opened 2022-05-11
Develop

DmytroSvirsa:


app/main.py
```diff
@@ -1,13 +1,25 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # white your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed"
+        if [format_linter_error(error) for error in errors] == []
```
MasterpieceElbow, line 16: You don't have to call a function for every error here. You have only to check if there are any errors or not.

MasterpieceElbow (changes requested):

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #55, opened 2022-05-12
Solution

vasylhnatiuk:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "name": error["code"],
+        "source": "flake8",
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # white your code here
-    pass
+    return {
+        "errors": [format_linter_error(errors[i]) for i in range(len(errors))],
```
MasterpieceElbow, line 13: `i` is not a good variable name here, `index` is probably better. But, actually, you can iterate just over the `errors`.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):

PR #66, opened 2022-05-13
'Solution'

YuraBeseda:


app/main.py
```diff
@@ -1,13 +1,46 @@
...
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8",
+            }
+            for error in errors
+        ],
```
TPolina, line 22: you should use the function you previously wrote

app/main.py
```diff
@@ -1,13 +1,46 @@
...
+                    "message": report[i]["text"],
+                    "name": report[i]["code"],
+                    "source": "flake8",
+                }
+                for i in range(len(report))
+            ],
+            "path": key,
+            "status": "passed" if len(report) == 0 else "failed",
+        }
+        for key, report in linter_report.items()
```
TPolina, line 44: same here

TPolina (changes requested):

TPolina (changes requested):

PR #68, opened 2022-05-19
'Solution'

LaskoA:


app/main.py
```diff
@@ -1,13 +1,22 @@
...
+            }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # white your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "passed" if not [format_linter_error(error)
+                                       for error in errors]else "failed"
```
Abnormaltype, line 14: It's better to write like this:
`"status": "failed" if errors else "passed"`

Abnormaltype (changes requested):

app/main.py
LaskoA: Done, please check again.

Abnormaltype (approved):
Well done!

PR #131, opened 2022-08-12
solution

yevhenii-nevmyvako:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "path": file_path,
+        "status": "passed" if errors == [] else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # white your code here
-    pass
+    return [format_single_linter_file(key, value)
+            for key, value in linter_report.items()]
```
Abnormaltype, line 21: what are key and value?

Abnormaltype (changes requested):

app/main.py
yevhenii-nevmyvako: key - it is the name of the dictionary list. like("./source_code_2.py": [])
value - it is the each key: value  in the dict. like ( "line": 112, "column": 6, "message": "no newline at end of file", "name": "W292", "source": "flake8",)

Abnormaltype: your names are uninformative - better use `file_path` and `errors`

yevhenii-nevmyvako: yes with these names I can't explain the answer

yevhenii-nevmyvako: half an hour I thought how to explain these names

Abnormaltype: make commit please

Abnormaltype: you don't need to explain these names
you should refactor your code
```
    return [format_single_linter_file(key, value)
            for file_path, errors in linter_report.items()]
```

yevhenii-nevmyvako: i made new commit 
'replaced the names in (def format_linter_report()) key - faile_path, value - errors'

Abnormaltype (approved):
Well done!

app/main.py
yevhenii-nevmyvako: Thank you, I will work on my mistakes

PR #148, opened 2022-08-25
wrote a three function which return

LitvinOleksandr:
1 collect information from error list and make a line information line, column, message, name, sorce
2 called first function and make dictionary with information and add file name which has error and status "passed" or "failed".
3 create a dictionary with all eroors, and show file whose has error

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # white your code here
-    pass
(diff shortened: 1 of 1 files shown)


Abnormaltype (changes requested):
Good job! but use `""` instead of `''`

Abnormaltype (approved):
Well done!

PR #188, opened 2022-09-13
Develop

alexzubkoff:


app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors":
```
Nattalli, line 12: your solution looks good but your code style doesn't. Read this checklist to fix errors:
In this line, it is the rule that `The curly braces must be located in one of two options: open and start with the text or have a line break between the text.`

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+            "path": file_path,
+            "status": "failed" if errors else "passed"
+            }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(key, linter_report[key])
+            for key in linter_report.keys()
```
Nattalli, line 21: and here you have to use descriptive and correct variable names. Rename your `key` variable
**Notice** if you iterate on linter_report it will return keys without specifying by method `.keys`

Nattalli: that means that 
```python
for key in linter_report.keys()
```
is the same as
```
for key in linter_report
```
so you can make our code easier for reading

Nattalli (changes requested):

Nattalli (approved):
Well done! 👍🏻

PR #195, opened 2022-09-13
Solution

graandda:


app/main.py
```diff
@@ -1,13 +1,28 @@
...
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
```
Nattalli, line 13: please, use code style as mentioned in this checklist:
Here check this rule:
`The curly braces must be located in one of two options: open and start with the text or have a line break between the text.`

app/main.py
```diff
@@ -1,13 +1,28 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [{"errors": [],
+             "path": f"{key}",
+             "status": "passed"}
+            if value == [] else {
```
Nattalli, line 24: look, you have the same "path" in both cases. also, if you have no errors - the `format_single_linter_file` function will return `[]` so you can create `errors` and `path` keys without the `if else` statement. Use `if else` only for `status` key

Nattalli (changes requested):

app/main.py
```diff
@@ -1,13 +1,22 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [{"errors": [format_linter_error(error) for error in value],
+             "path": f"{key}",
+             "status": "passed" if linter_report[key] == [] else "failed"}
+            for key, value in linter_report.items()]
```
Nattalli, line 22: everything looks good now but use descriptive and correct variable names(instead of `key` and `value` you can use file_path and errors or something like that)

Nattalli (approved):
Cool! 👍🏻

PR #207, opened 2022-09-14
solved the task

Cerne13:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(path, error)
+        for path, error in linter_report.items()
```
MasterpieceElbow, line 22: `error` means a single object. But linter_report values are lists.

MasterpieceElbow (changes requested):

app/main.py
Cerne13: Thank you!
Yes, it's a silly mistake. Of course we pass a list of errors to the function.

Cerne13:
Hi! Could you re-review my task please? 
I've changed the variable name to 'errors' since it's a list and pushed a new commit.
Thank you.

Nattalli (approved):
👍🏻

PR #208, opened 2022-09-14
linter attempt 1

VitaliyBychevoy:


app/main.py
```diff
@@ -1,13 +1,25 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
MasterpieceElbow, line 24: It is better to use a name what it means when you pass it to `format_single_linter_file`. So, `format_single_linter_file(key, value)` is not very clear about what arguments you pass to the function.

MasterpieceElbow (changes requested):

MasterpieceElbow (approved):
GJ!

PR #213, opened 2022-09-14
'Solution'

serviza:


app/main.py
```diff
@@ -1,13 +1,20 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
```
Nattalli, line 2: your code looks good but the curly braces must be located in one of two options: open and start with the text or have a line break between the text.
You can see examples here:

Nattalli (changes requested):

app/main.py
serviza: Ok! Solved! )

Nattalli (approved):
👍🏻

PR #225, opened 2022-09-16
Solution

AndriyKy:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
Abnormaltype, line 22: key and value are uninformative names for a variables

Abnormaltype (changes requested):

app/main.py
AndriyKy: Key and value names was changed

AndriyKy: Key and value names was changed

Abnormaltype (approved):
Well done!

PR #236, opened 2022-09-25
Tested solution

Ida83:


app/main.py
```diff
@@ -1,13 +1,17 @@
...
+            if len(errors) > 0 else "passed"
+            }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(path, linter_report[path])
+            for path in linter_report
+            ]
```
DmytroSvirsa, line 17: read checklist.md to correct your code style

DmytroSvirsa (changes requested):

app/main.py
Ida83: I don't understand what is wrong.
return [
            format_single_linter_file(path, linter_report[path])
            for path in linter_report
            ]
such code call flake8 error

app/main.py
```diff
@@ -1,13 +1,18 @@
...
+            }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(i) for i in errors],
+            "path": file_path, "status": "failed"
+            if len(errors) > 0 else "passed"
+            }
```
DmytroSvirsa, line 11: correct for dictionary also

app/main.py
```diff
@@ -1,13 +1,18 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {"line": error["line_number"], "column": error["column_number"],
+            "message": error["text"], "name": error["code"], "source": "flake8"
+            }
```
DmytroSvirsa, line 4: correct for dictionary also

DmytroSvirsa (changes requested):

app/main.py
Ida83: done

DmytroSvirsa (approved):
Well done!

PR #255, opened 2022-10-07
'Solution'

contritum:


app/main.py
```diff
@@ -1,13 +1,45 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors":
+            [
+                {
```
IvanRamyk, line 15: use function `format_linter_error` to format single error. It allows you to avoid code duplication

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+        "status": "failed" if len(errors) > 1 else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "path": keys,
```
IvanRamyk, line 32: use function `format_single_linter_file ` to format a single file. It allows you to avoid code duplication

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #271, opened 2022-10-21
solution of task

phaishuk:
seems to me all okay   :)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,23 @@
+
+
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error.get("line_number"),
+        "column": error.get("column_number"),
+        "message": error.get("text"),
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "name": error.get("code"),
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(value) for value in errors],
```
Abnormaltype, line 15: value it's a bad name for a variable

Abnormaltype (changes requested):

app/main.py
phaishuk: variable name changed to "error" in 15th row

Abnormaltype (approved):
Well done!

PR #290, opened 2022-11-05
'Solution'

Sergey-Bushinskiy:


Sergey-Bushinskiy:
delete empty line in the code end.

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "status": "failed" if any(errors) else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(path_name, list_of_errors)
+            for path_name, list_of_errors in linter_report.items()
+            ]
```
DmytroSvirsa, line 22: make code style better here (read the checklist, but I think you know what is wrong)

DmytroSvirsa (changes requested):

app/main.py
Sergey-Bushinskiy: I corrected it, watsh please.

Sergey-Bushinskiy: I corrected it, watch please.

Sergey-Bushinskiy: @DmytroSvirsa  i  corrected last mistake.

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(path_name, list_of_errors)
+        for path_name, list_of_errors in linter_report.items()
+    ]
```
YuriiKindrat, line 23: It is better to break the line with generators in such way:
[
some variable
for some variable
in some iterable
]

YuriiKindrat (changes requested):

app/main.py
Sergey-Bushinskiy: @YuriiKindrat     return [
        format_single_linter_file(path_name, list_of_errors)
        for path_name, list_of_errors in linter_report.items()
    ]   in this variant there are no mistakes  in flake8 app, if i use your variant , i  have 2 mistakes, so i did another way. Watch please.

PR #315, opened 2022-11-20
'Solution'

Serhii-Stolitnii:


Abnormaltype (changes requested):

app/main.py
```diff
@@ -1,13 +1,24 @@
...
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors":
+            [
+                format_linter_error(error) for error in errors
+            ],
+        "path": str(file_path),
```
YuriiKindrat, line 15: No need to use str() here, according to type annotation it is already str format

YuriiKindrat (changes requested):

app/main.py
Serhii-Stolitnii: fixed

app/main.py
```diff
@@ -1,13 +1,24 @@
...
+        "status": "passed" if len(errors) == 0 else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, errors)
+        for key, errors in linter_report.items()
```
Abnormaltype, line 23: key - uninformative name for a variable

Abnormaltype (changes requested):

YuriiKindrat (approved):

app/main.py
Serhii-Stolitnii: fixed

PR #327, opened 2022-12-04
Solution

TarasSavchyn:


DmytroSvirsa (changes requested):
generaly good job, but remove .pyc files from commit

PR #328, opened 2022-12-05
Solution

Rainmanuk:


app/main.py
```diff
@@ -1,13 +1,30 @@
...
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": [],
+            "path": error,
+            "status": "passed"
+        }
+        if linter_report[error] == []
```
YuriiKindrat, line 26: no need to make this condidtiton here, make check for status in format_single_linter_file function in "status" key

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,13 +1,25 @@
...
+        )
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(error, linter_report[error])
+        for error in linter_report
```
Serhii-Leonenko, line 24: It's better to use dict.items() here, because you use both key and value, and then you can use value instead of dict[key]

app/main.py
```diff
@@ -1,13 +1,25 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": (
+            "failed" if errors != [] else "passed"
```
Serhii-Leonenko, line 16:
```suggestion
            "failed" if errors  else "passed"
```

You can remove it,  and check if errors are not empty

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,13 +1,25 @@
...
+        )
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
Serhii-Leonenko, line 24: Use here more descriptive names for variables instead of key and value.

Serhii-Leonenko (changes requested):

YuriiKindrat (approved):

PR #339, opened 2022-12-13
Solution

OleksandrGrk:


app/main.py
```diff
@@ -1,13 +1,17 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {"line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"}
```
IvanRamyk, line 6:
```suggestion
    return {
        "line": error["line_number"],
        "column": error["column_number"],
        ...
    }
```

IvanRamyk: I believe this way of formatting looks better 
could you fix everywhere?

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #341, opened 2022-12-16
Solution

Tania-Kharchuk:


IvanRamyk (changes requested):

app/main.py
```diff
@@ -1,13 +1,27 @@
...
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [
+        {"line": error["line_number"], "column": error["column_number"],
```
katevasilenko, line 11: Please, use your function `format_linter_error` here. Now you are repeat yourself

app/main.py
```diff
@@ -1,13 +1,27 @@
...
+    ],
+        "path": file_path,
+        "status": "failed" if errors else "passed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [{"errors": [
+        {"line": error["line_number"], "column": error["column_number"],
```
katevasilenko, line 21: And here too `format_single_linter_file ` DRY - don’t repeat yourself

katevasilenko (changes requested):
Nice, just use your own functions to avoid self repeating

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
Abnormaltype, line 22: key and value are uninformative names for a variables

Abnormaltype (changes requested):

app/main.py
Tania-Kharchuk: Складається враження, що ви доскіпуєтесь. Маю велику надію, що це враження хибне.

katevasilenko (approved):

app/main.py
katevasilenko: Using informative variables is very important thing in developing. Your code will read another developers, they will should understand your code quick and easy. The good informative naming makes it possible. It’s better to learn this way from the very begging

Tania-Kharchuk: Got it, thanks

PR #345, opened 2022-12-17
Solution

numerodiciannove:
1. function format_linter_error (✓done)
2. function format_single_linter_file (✓done)
3. function format_linter_report (✓done)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -18,6 +18,6 @@ def format_single_linter_file(file_path: str, errors: list) -> dict:
 
 def format_linter_report(linter_report: dict) -> list:
     return [
-        format_single_linter_file(flake8_errors, linter_report[flake8_errors])
-        for flake8_errors in linter_report
+        format_single_linter_file(key, value)
+        for (key, value) in linter_report.items()
     ]
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,22 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors != [] else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(flake8_errors, linter_report[flake8_errors])
+        for flake8_errors in linter_report
```
Serhii-Leonenko, line 22: It's better to use linter_report.items() here, because you use both key and value.

Serhii-Leonenko (changes requested):

app/main.py
numerodiciannove: done

katevasilenko (approved):
Looks good to me

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors != [] else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for (key, value) in linter_report.items()
```
andrewbelo, line 22: key, value не годиться, назви їх якось

andrewbelo (changes requested):

app/main.py
numerodiciannove: обновил

Abnormaltype (approved):

PR #353, opened 2022-12-22
Solution

ilushasm:


app/main.py
```diff
@@ -1,13 +1,27 @@
...
+        "status": "passed" if len(errors) < 1 else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, linter_report[key])
+        for key in linter_report
```
katevasilenko, line 26: Please use key and value here and give them informative names

katevasilenko (changes requested):

app/main.py
ilushasm: done

katevasilenko (approved):
Good job

PR #356, opened 2022-12-28
linker formatter done

VadymRabotiuk:


app/main.py
```diff
@@ -1,13 +1,19 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {"line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"
+            }
```
IvanRamyk, line 7:
```suggestion
    return {
        "line": error["line_number"],
        "column": error["column_number"],
        "message": error["text"],
        "name": error["code"],
        "source": "flake8"
    }
```

IvanRamyk: I prefer different formatting and it's more common among python developers

IvanRamyk: please, fix it everywhere

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #386, opened 2023-01-24
Solution

mkobylianskyi:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(i) for i in errors],
+        "path": file_path,
+        "status": ("passed") if errors == [] else ("failed")
```
Bohdan-Salamakha, line 15: redundant parenthesis
not `("passed")` but "passed"

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(i) for i in errors],
```
Bohdan-Salamakha, line 13: I think better will be not `i` but `error`

app/main.py
Bohdan-Salamakha, line 15: `if not errors`

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": ("passed") if errors == [] else ("failed")
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, linter_report[key])
+        for key in linter_report
```
Bohdan-Salamakha, line 22: not `key` but `file_report`
It's better for understanding  if u are a newbie on the project.
think about what if u will read your code for first time and u don't know anything about func logic (it's written by another dev)?
I think it's hard to understand what `key` means, especially in a big code fragments.

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):
great

PR #389, opened 2023-01-26
with list coprehension

flatchers:


app/main.py
```diff
@@ -1,13 +1,10 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {"line": error["line_number"],"column": error["column_number"], "message": error["text"], "name": error["code"], "source": "flake8"}
```
ArsenPidhoretskyi, line 2: Please format as follow:
```
return {
    "key": value,
    "key1": value
}
```

ArsenPidhoretskyi: Also in rest places, so hard to read

ArsenPidhoretskyi (changes requested):
Fix flake8

Vasyl-Smutok (approved):
Well Done 👍

PR #391, opened 2023-01-27
Solution

Devrro:
That was fun!Can`t wait to win!


thorium-zero (approved):
Clean and precise. Great solution

PR #397, opened 2023-01-30
Solution

ZabFTFT:


app/main.py
```diff
@@ -1,13 +1,20 @@
...
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": ["failed" if errors else "passed"][0]}
```
wQuelS, line 14: do you really need an additional list here as a value?

wQuelS (changes requested):

app/main.py
ZabFTFT: I didn't know how to pass if else without a list =) And now I know that it's real. Thank you!

wQuelS (approved):
Nice

PR #398, opened 2023-01-30
Dev roman senyshyn py linter formatter 

AndereLion:


app/main.py
```diff
@@ -1,13 +1,25 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(
+            file, linter_report[file]
+        ) for file in linter_report
```
wQuelS, line 24: everything that inside return list is formatted in pretty strange way. You may use this tips to follow best practices: https://github.com/mate-academy/py-linter-formatter/blob/master/checklist.md

Or you can even use black package to make it automatically, docs: https://pypi.org/project/black/

wQuelS (changes requested):

app/main.py
AndereLion: Fixed!

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
-def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+def format_linter_report(lin_rep: dict) -> list:
+    return \
+        [format_single_linter_file(file, lin_rep[file]) for file in lin_rep]
```
YuriiKindrat, line 21: try not to use slash here, it is better to make this way, but with list
image

YuriiKindrat (changes requested):

app/main.py
AndereLion: Fixed (without slash) .

YuriiKindrat (approved):

PR #409, opened 2023-01-31
Solution

usrofgh:


app/main.py
```diff
@@ -1,13 +1,27 @@
...
+        "source": "flake8",
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors":
+            [
```
YuriiKindrat, line 14: move this bracket near the dict key, for better code style

app/main.py
```diff
@@ -1,13 +1,27 @@
...
+                format_linter_error(error_iter)
+                for error_iter in errors
+            ],
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
```
YuriiKindrat, line 19: u can write here just "if errors" no need to use len method here

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #416, opened 2023-02-01
Solution for linter.

DmytryTong:


app/main.py
```diff
@@ -1,13 +1,31 @@
...
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed"
+    } if errors else \
+        {
+        "errors": [],
+        "path": file_path,
+        "status": "passed"
+    }
```
YuriiKindrat, line 23: no need to make other dict for return with status passed u can make this condition in "status" key with passed/failed words

YuriiKindrat (changes requested):

DmytryTong:
Soved problem, please checkout.

PR #417, opened 2023-02-01
Solution

Yevheniy-Lototskiy:


YuriiKindrat (changes requested):
Please, remove __pass__ keyword from every function

YuriiKindrat (approved):

PR #421, opened 2023-02-01
Solution

oksana0410:


app/main.py
```diff
@@ -1,13 +1,25 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(err) for err in errors],
+        "path": file_path,
+        "status": "failed"
+        if [format_linter_error(err) for err in errors]
```
YuriiKindrat, line 16: no need to iterate through errors for second time,f.e. u can check the len of errors list

app/main.py
```diff
@@ -1,13 +1,25 @@
...
+        else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(item, linter_report[item])
+        for item in linter_report
```
YuriiKindrat, line 24: u can iterate through linter_report.items(), so it won't need to call linter_report[item] in 23 line

YuriiKindrat (dismissed):

app/main.py
oksana0410: don't understand how to take this key value to check len without dict name

app/main.py
```diff
@@ -1,13 +1,24 @@
...
+                  else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
wQuelS, line 23: it is better to use more descriptive names, instead of `key, value`. For example, here you have "path" and "errors"

wQuelS (approved):
Cool

PR #423, opened 2023-02-01
Solution

oigod:


YuriiKindrat (changes requested):
Please, remove pass keyword from every function

PR #424, opened 2023-02-01
solution

Yulia-Mala:


app/main.py
```diff
@@ -1,13 +1,14 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {"source": "flake8", "line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"], "name": error["code"]}
```
YuriiKindrat, line 4: image
fix this according to the best practicies, use line only for one key/value pair

app/main.py
```diff
@@ -1,13 +1,14 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors]} | \
+           {"path": file_path, "status": "failed" if errors else "passed"}
```
YuriiKindrat, line 9: the same here

YuriiKindrat (changes requested):

Yulia-Mala:
I have fixed it, thanks!

app/main.py
```diff
@@ -1,13 +1,17 @@
...
+            "message": error["text"],
+            "name": error["code"]}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors]} | \
+           {"path": file_path,
+            "status": "failed" if errors else "passed"}
```
YuriiKindrat, line 12: why do u need 2 dicts here, u can use only one, with all needed arguments

YuriiKindrat (changes requested):

Nattalli (approved):
Good job!

PR #429, opened 2023-02-01
Solution

Anatolii-Poznyak:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(errors[i]) for i in range(len(errors))],
```
YuriiKindrat, line 13: no need to use range(len()) here u can just iterate through errors

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #433, opened 2023-02-01
Solution

ziriael-absten:


app/main.py
```diff
@@ -1,13 +1,19 @@
...
+            "path": file_path,
+            "status": "failed" if len(errors) != 0 else "passed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(path, errors)
+        for (path, errors) in linter_report.items()
```
wQuelS, line 18: You don't need curly braces when iterating key, value pair

wQuelS (changes requested):

wQuelS (approved):
Nice job

PR #440, opened 2023-02-01
Solution

tytomyr:


.gitignore
```diff
@@ -1,7 +1,10 @@
 .idea/
 .vscode/
+*.pyc
 *.iml
 .env
 .DS_Store
 venv/
 .pytest_cache/
+__pycache__/
+*.py[cod]
```
wQuelS, line 10: If you decided to add gitignore, do it properly. Add a new line at the end of file

app/main.py
```diff
@@ -1,13 +1,28 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [
+            format_linter_error(error) for error in errors
+        ],
```
wQuelS, line 17: this key-value pair could go in one line, since it is not too long. It is up to you to decide whether to change it or not in this case. So I will not reject if you decide leave it as it is.

wQuelS (changes requested):

.gitignore
tytomyr: added a new line at the end

YuriiKindrat (approved):

PR #442, opened 2023-02-01
Solution

me1nyk:


app/main.py
```diff
@@ -1,13 +1,28 @@
...
+        "status": "failed" if len(errors) > 0 else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": [format_linter_error(error)
```
wQuelS, line 22: you don't need to duplicate this chunk of code, because you can use function that u've written above

wQuelS (changes requested):

app/main.py
```diff
@@ -1,13 +1,24 @@
...
+        "status": "failed" if len(errors) > 0 else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path=path,
+                                  errors=errors)
```
wQuelS, line 22: you don't actually need key arguments, you can pass them both as usual positional ones

wQuelS (changes requested):

app/main.py
me1nyk: it just looks better in my opinion, thanks for checking:)

wQuelS (approved):
Good job

Serhii-Leonenko (approved):
Good job!

PR #445, opened 2023-02-01
Solution

VladyslavCherkashyn:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(dictionary, linter_report[dictionary])
+        for dictionary in linter_report
```
YuriiKindrat, line 22: u can iterate here through linter_report.items(), so there won't need to take value by a key

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #448, opened 2023-02-01
'Solution'

dogmoony:
I have a problem: command "pytest" isn`t working. Maybe I`m doing something wrong, but still. What could it be?

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,17 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {"line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["physical_line"],
+            "source": error["store and decorate numbers"]}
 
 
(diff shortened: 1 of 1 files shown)


YuriiKindrat (changes requested):
All tests should pass - fix ur solution

app/main.py
```diff
@@ -1,13 +1,17 @@
...
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "failed" if errors else "passed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(element[0], element[1])
```
wQuelS, line 16: why? you could just iterate with, for example, path and errors over dict.items()

wQuelS (changes requested):

wQuelS (approved):
Well done

PR #456, opened 2023-02-03
'Solution'

ibulakevych:


app/main.py
```diff
@@ -1,13 +1,27 @@
...
+        "status": "failed" if len(errors) > 0 else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path, linter_report[file_path])
+        for file_path in list(linter_report.keys())
```
ArsenPidhoretskyi, line 26: you can use .items  on dict
for file_path, errors in linter_report.items()

ArsenPidhoretskyi (changes requested):

app/main.py
ibulakevych: Ok, I've changed it

ibulakevych: Can you check the new variant?

olena-hul (approved):

olena-hul:
Good job!

PR #463, opened 2023-02-07
Solution

yhwh6:


app/main.py
```diff
@@ -1,13 +1,30 @@
...
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error)
+            for error in errors
+        ],
+        "path": file_path,
+        "status": check_status(errors)
```
katevasilenko, line 18: Why just don’t check right here. This part of code doesn’t repeated any where

katevasilenko (changes requested):

Vasyl-Smutok (approved):
Good Job 👍

PR #464, opened 2023-02-09
Solution

oleg-shcherbak:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
```
Bohdan-Salamakha, line 11: list of what?
add detailed typehint

app/main.py
```diff
@@ -1,13 +1,23 @@
...
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
```
Bohdan-Salamakha, line 19: list of what?
add detailed typehint

Bohdan-Salamakha (changes requested):
v1

app/main.py
oleg-shcherbak: I should do that?
errors: list[dict]

PR #476, opened 2023-02-18
Solution

AlyonkaB:


app/main.py
```diff
@@ -1,13 +1,26 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
ArsenPidhoretskyi, line 25: key, value is not clear names

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,13 +1,26 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [format_linter_error(i) for i in errors],
```
katevasilenko, line 15: Also `i` is not good variable in this case. `i` Good for indexes. Please use `error` here

katevasilenko (changes requested):

katevasilenko (approved):
Well done!

PR #480, opened 2023-02-21
task done

chipapapo:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return \
+        [format_single_linter_file(i, linter_report[i]) for i in linter_report]
```
ArsenPidhoretskyi, line 21: First of all
```suggestion
    return [
        format_single_linter_file(i, linter_report[i])
        for i in linter_report
    ]
```

ArsenPidhoretskyi: Also, use linter_report.items()
and use good names `i` is not good variable name

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #511, opened 2023-03-20
Solution

alim2709:


app/main.py
```diff
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```
Bohdan-Salamakha, line 6: use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
```
Bohdan-Salamakha, line 11: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

app/main.py
```diff
@@ -1,13 +1,23 @@
...
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": ("failed" if errors != [] else "passed")
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
```
Bohdan-Salamakha, line 19: add additional typehint for list inner objects like list[YOUR_VALUE_TYPE]
for example: list that contain string elements has typehint list[str]

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error.get("line_number", None),
+        "column": error.get("column_number", None),
+        "message": error.get("text", None),
+        "name": error.get("code", None),
```
Bohdan-Salamakha, line 6: u don't need to set up default  None values

app/main.py
```diff
@@ -1,13 +1,23 @@
...
 
 
-def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+def format_single_linter_file(file_path: str, errors: list[dict]) -> dict:
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if errors != [] else "passed"
```
Bohdan-Salamakha, line 15: `"status": "failed" if errors else "passed"`

Bohdan-Salamakha (changes requested):
v1

app/main.py
alim2709: я могу тут с вами обсуждать если что то не понятно?

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
-def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+def format_linter_report(linter_report: dict) -> list[dict]:
+    return [
+        format_single_linter_file(key, linter_report.get(key, None))
```
Bohdan-Salamakha, line 21: u don't need to set up None default value
        format_single_linter_file(key, linter_report.get(key, None))

Bohdan-Salamakha (changes requested):
v2

Bohdan-Salamakha (approved):

PR #513, opened 2023-03-22
Solution py-linter-formatter

vborovets0:


app/main.py
```diff
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```
Bohdan-Salamakha, line 6: use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
```
Bohdan-Salamakha, line 15:
```suggestion
        "status": "failed" if errors else "passed"
```

olena-hul:
Good job!

PR #522, opened 2023-04-01
Solution

bvzakora:


app/main.py
```diff
@@ -1,13 +1,30 @@
...
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status":
+            "failed" if [format_linter_error(error) for error in errors]
+            else "passed"
```
EdAlekseiev, line 18: Don't format `errors` twice, in this case, you must check `errors` is empty or not

EdAlekseiev (changes requested):

app/main.py
bvzakora: Fixed error checking

PR #523, opened 2023-04-01
Create initial implementation of the module

barbequeue:


tests/test_main.py
```diff
@@ -85,14 +85,26 @@ def test_format_linter_error(error_linter, error_mate):
...
 def test_format_functions_one_line(func):
     code = inspect.getsource(func)
+    docstring = ast.get_docstring(ast.parse(code).body[0])
     assert (
-        isinstance(ast.parse(code).body[0].body[0], ast.Return) is True
-    ), f"Function '{func.__name__}' should contain only return statement"
+        isinstance(docstring, str)
+        and isinstance(ast.parse(code).body[0].body[1], ast.Return) is True
+        or isinstance(ast.parse(code).body[0].body[0], ast.Return) is True
+    ), f"Function '{func.__name__}' should contain only return statement and optional docstring"
```
barbequeue, line 93: This test condition doesn't allow me to add a function docstrings to my module, so i change this condition to allow optional function docstrings in my modile

Bohdan-Salamakha: u don't need it

Bohdan-Salamakha (changes requested):
fix tests

tests/test_main.py
barbequeue: Got it, i just wanted to add the docstrings

app/main.py
```diff
@@ -1,13 +1,29 @@
...
+
+
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```
Bohdan-Salamakha, line 12: it's preferable to use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError

app/main.py
```diff
@@ -1,13 +1,29 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
```
Bohdan-Salamakha, line 21:
```suggestion
        "status": "failed" if errors else "passed"
```

Bohdan-Salamakha (changes requested):
v1

Bohdan-Salamakha (approved):

PR #532, opened 2023-04-11
Solution

KittNeKit:


app/main.py
```diff
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```

Bohdan-Salamakha (changes requested):
v1

PR #535, opened 2023-04-11
Develop

sind14:
Наче змінив на develop

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -18,6 +18,6 @@ def format_single_linter_file(file_path: str, errors: list) -> dict:
 
 def format_linter_report(linter_report: dict) -> list:
     return [
-        format_single_linter_file(file_path=dict_key, errors=dict_value)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if len(errors) > 0 else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path=dict_key, errors=dict_value)
+        for dict_key, dict_value in linter_report.items()
```
ArsenPidhoretskyi, line 22: dict_key, dict_value is not clear names again...
Rename dict_key to file_path, dict_value to errors

ArsenPidhoretskyi (changes requested):

app/main.py
sind14: ok)

PR #542, opened 2023-04-18
Solution

notrealoxx:


.flake8
ArsenPidhoretskyi, line 1: What you change here?
The same with test.yml, .gitignore and others
Check this link, only main.py must changed 
https://github.com/mate-academy/py-linter-formatter/pull/542/files

app/main.py
```diff
@@ -1,13 +1,44 @@
...
+    text = error.get("text", "")
+    physical_line = error.get("physical_line", "")
+
+    formatted_error = f"Code: {code}\n"
+    formatted_error += f"Filename: {filename}\n"
+    formatted_error += f"Line number: {line_number}\n"
+    formatted_error += f"Column number: {column_number}\n"
+    formatted_error += f"Text: {text}\n"
+    formatted_error += f"Physical line: {physical_line}\n"
+    return formatted_error
```
ArsenPidhoretskyi, line 16: It must return dict not string.

ArsenPidhoretskyi (changes requested):
The task not work good. Please use list comprehension!
Also pull request must be from develop to master. Not from master to master

PR #553, opened 2023-04-27
List and Dict Comprehensions solution

FadieievaA:


app/main.py
```diff
@@ -1,13 +1,22 @@
...
+            "path": file_path,
+            "status": ("passed" if len(errors) == 0 else "failed")}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path=key, errors=value)
+        for key, value in linter_report.items()
```
ArsenPidhoretskyi, line 20: key, value is not clear names
Please choose more better names

ArsenPidhoretskyi (changes requested):

PR #562, opened 2023-05-07
Solution

serji13:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"path": file_path,
+            "status": "passed",
+            "errors": []} if errors == [] else \
+        {"path": file_path,
+         "status": "failed",
+         "errors": [format_linter_error(i) for i in errors]
+         }
```
ArsenPidhoretskyi, line 19: So hard... You can just add if else only to "status"
```
{
    "path": file_path,
     "status": "failed" if errors else "passed",
     "errors": [format_linter_error(error) for error in errors]
}
```

app/main.py
```diff
@@ -1,13 +1,23 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(i, n) for i, n in linter_report.items()]
```
ArsenPidhoretskyi, line 23: i, n is not clear names. Please choose better names

app/main.py
```diff
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return \
+        {
```
ArsenPidhoretskyi, line 3:
```suggestion
    return {
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "path": file_path,
+        "status": "failed" if errors else "passed",
+        "errors": [format_linter_error(i) for i in errors]
```
Bohdan-Salamakha, line 15: what is "i"?
it's preferable to give understandable names to variables
use "error" instead of "i"

Bohdan-Salamakha (changes requested):
v1

app/main.py
serji13: sorry, my bad

Bohdan-Salamakha (approved):

PR #567, opened 2023-05-09
03_linter_formatter

ruslanvd:


app/main.py
```diff
@@ -1,13 +1,26 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [
+        format_single_linter_file(file_path=x, errors=y)
+        for x, y in linter_report.items()
```
ArsenPidhoretskyi, line 25: x and y is not clear names
Please choose better names for this

ArsenPidhoretskyi (changes requested):

app/main.py
ruslanvd: fixed :)

ArsenPidhoretskyi (approved):
Well done!

PR #589, opened 2023-05-16
Implemented Linter formatter tasks

OleksiiKiva:


app/main.py
```diff
@@ -1,13 +1,46 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```

app/main.py
```diff
@@ -1,13 +1,46 @@
...
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8"
+                }
+                for error in errors
+            ],
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
+    }
```
Bohdan-Salamakha, line 26: u need to use format_linter_error() here to avoid code duplication

app/main.py
```diff
@@ -1,13 +1,46 @@
...
+                        "message": value["text"],
+                        "name": value["code"],
+                        "source": "flake8"
+                    }
+                    for value in linter_report[report]],
+            "path": report,
+            "status": "failed" if len(linter_report[report]) > 0 else "passed"
+        }
+        for report in linter_report
+    ]
```
Bohdan-Salamakha, line 46: u need to use format_single_linter_file() here to avoid code duplication

Bohdan-Salamakha (changes requested):
v1

app/main.py
OleksiiKiva: Fixed, I think

app/main.py
OleksiiKiva: Done, maybe

app/main.py
OleksiiKiva: Done

Bohdan-Salamakha (changes requested):
v2

Bohdan-Salamakha (approved):

PR #591, opened 2023-05-17
Solution

AlexIzmashkin:


app/main.py
```diff
@@ -1,13 +1,41 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```

app/main.py
```diff
@@ -1,13 +1,41 @@
...
+            "line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"
+        }
+        for error in errors],
+        "path": file_path,
+        "status": "failed"
+    }
```
Bohdan-Salamakha, line 23: use format_linter_error() function here

app/main.py
AlexIzmashkin: why mentor didn't use get() method in this case? 
image

Bohdan-Salamakha (approved):

PR #593, opened 2023-05-22
Develop

Feristis:


app/main.py
```diff
@@ -1,13 +1,27 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```

app/main.py
```diff
@@ -1,13 +1,27 @@
...
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"path": file_path, "status": "failed" if any(errors) else "passed",
+            "errors": [{"line": error["line_number"],
+                        "column": error["column_number"],
+                        "message": error["text"],
+                        "name": error["code"],
+                        "source": "flake8"}for error in errors]}
```
Bohdan-Salamakha, line 17: use format_linter_error() function here

Bohdan-Salamakha (changes requested):

PR #594, opened 2023-05-22
done

maccarets:


app/main.py
```diff
@@ -1,13 +1,40 @@
...
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [{
+            "line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"
```
Oleg-s-s, line 18: Same lines as in 3-7, think about using that function in this function

app/main.py
```diff
@@ -1,13 +1,40 @@
...
+            "errors": [{
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
+            } for error in errors],
+
+            "path": key,
+            "status": "failed" if len(errors) > 0 else "passed"
```
Oleg-s-s, line 38: Same here. You can even try to use the updated second function here.

LiudmylaKulinchenko (approved):
Nice work! 🎉

PR #595, opened 2023-05-22
Solution

YevheniiIovchu:
In the last function I compare **if** [](value one of the items of incoming dict) is **not report**  - is not empty (== True) and this way I generate the part of the resulting dict where "status": "passed". In the second part of incoming dict("error" is the name of each incoming dict, whose value not []) by comparison **if report** I generate the remaining resulting dict.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,50 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
+            } for error in errors
+        ],
+        "path": file_path,
+        "status": "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [
+        {
+            "errors": report,
+            "path": path,
+            "status": "passed"
+        } if not report
+        else {
+            "errors": [
+                {
+                    "line": error["line_number"],
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8"
+                } for error in report if report
+            ],
+            "path": path,
+            "status": "failed"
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,50 @@
...
+    return {
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
+            } for error in errors
+        ],
```
YuriiKindrat, line 23: follow DRY principle in your code - u can use function implemented before )

app/main.py
```diff
@@ -1,13 +1,50 @@
...
+        "path": file_path,
+        "status": "failed"
```
YuriiKindrat, line 25: image
fix this according to the task

app/main.py
```diff
@@ -1,13 +1,50 @@
...
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8"
+                } for error in report if report
+            ],
+            "path": path,
+            "status": "failed"
+        } for path, report in linter_report.items()
+    ]
```
YuriiKindrat, line 50: same here, try to use functions implemented before

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -12,39 +12,17 @@ def format_linter_error(error: dict) -> dict:
...
-                    "column": error["column_number"],
-                    "message": error["text"],
-                    "name": error["code"],
-                    "source": "flake8"
-                } for error in report if report
-            ],
+            "errors": format_single_linter_file(path, report)["errors"],
             "path": path,
-            "status": "failed"
+            "status": "passed" if not report else "failed"
```
Oleg-s-s: You have path and status already implemented in format_single_linter_file. Think about handling reports in format_linter_report 1 by 1 using format_single_linter_file

Oleg-s-s (changes requested):

app/main.py
YevheniiIovchu: Do you mean I need to use in my third function params from my second function? And I should replace path -> file_path and report -> errors like here?
{
24 "errors": format_single_linter_file(file_path, errors)["errors"],
25 "path": file_path,
26 "status": "passed" if not errors else "failed"
27  for file_path, errors in linter_report.items()
}
Or my third function should accept data from my second function and it should define "path" and "status" by itself without any extra conditions like in this line 
26 "status": "passed" if not errors else "failed"
where I repeat the same action as in the second function.

YevheniiIovchu: And is it good practice to handle conversation here?

YuriiKindrat (approved):
GJ!

PR #601, opened 2023-05-23
"Solution"

Po-SvinSkipodla:


app/main.py
```diff
@@ -1,13 +1,26 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [format_linter_error(err) for err in errors],
```
Oleg-s-s, line 15: Try using more descriptive names of variables, error isn't such a big name to be shorten)

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,13 +1,26 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if len(errors) != 0 else "passed"
```
Oleg-s-s, line 17: No need to check the len() here, existence will be ok. And mark in the table for mentor review next time please

Oleg-s-s (changes requested):

PR #602, opened 2023-05-23
Solution

diana-shyrokikh:


app/main.py
```diff
@@ -1,13 +1,26 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file, info)
+        for file, info in linter_report.items()
```
YuriiKindrat, line 25: it is better to name info variable as errors, cause info is not descriptive name for var in that case

YuriiKindrat (approved):

PR #612, opened 2023-05-23
Add the execution result of flake8 errors

yevpy:


app/main.py
```diff
@@ -1,13 +1,19 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
+    return {"line": error["line_number"], "column": error["column_number"], "message": error["text"],
```
olena-hul, line 2: Fix flake8

olena-hul (changes requested):

app/main.py
yevpy: fixed

app/main.py
```diff
@@ -1,13 +1,31 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
     pass
```
YukihiroSM, line 9: pass should be removed, it's useless in this case. (remove it in all functions)

app/main.py
```diff
@@ -1,13 +1,31 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
```
YukihiroSM, line 12: list of what? Add more descriptive typehint, please. Like list[SOME_DATATYPE]

YukihiroSM (changes requested):

YukihiroSM (approved):
Looks good. Thanks for fixing issues.
For future you can also use even more descriptive type hints, using typing library. 
As example: 
`def format_linter_error(error: Dict[str, Any]) -> Dict[str, Any]`
But it depends on code readability, it's not always useful :)

PR #614, opened 2023-05-24
Done

arturiermolenko:
Implemented the solution for functions "format_linter_error", "format_single_linter_file " and "format_linter_report".

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,52 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors":
+            [
+                {
+                    "line": error["line_number"],
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8"
+                }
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,52 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```

app/main.py
```diff
@@ -1,13 +1,52 @@
...
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8"
+                }
+                for error in errors
+            ],
+        "path": file_path,
+        "status": "failed"
+    }
```
Bohdan-Salamakha, line 26: use format_linter_error() function here to avoid code duplication

app/main.py
```diff
@@ -1,13 +1,52 @@
...
+                    "name": error["code"],
+                    "source": "flake8"
+                }
+                for error in value
+            ],
+            "path": key,
+            "status": "failed"
+        }
+        for key, value in linter_report.items()
+    ]
```
Bohdan-Salamakha, line 52: use format_single_linter_file() function here to avoid code duplication

Bohdan-Salamakha (changes requested):

app/main.py
arturiermolenko: Thank You so much, this was good lesson for me!

olena-hul (approved):

olena-hul:
Good job!

PR #616, opened 2023-05-24
Solution

erikagayan:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(file_path, errors)
```
LiudmylaKulinchenko, line 20: The braces should be located in one of two options: open and start with the text or have a line break between the text. Please, use the same style through the code

app/main.py
```diff
@@ -18,4 +18,5 @@ def format_single_linter_file(file_path: str, errors: list) -> dict:
 
 def format_linter_report(linter_report: dict) -> list:
     return [format_single_linter_file(file_path, errors)
```
LiudmylaKulinchenko, line 20: Fix indentation according to the best practices

LiudmylaKulinchenko (changes requested):

app/main.py
erikagayan: Sorry, but I still don't understand what to fix. flake8 doesn't write anything. And when I try to write everything in one line, it gives an error: app\main.py:20:80: E501 line too long (104 > 79 characters)

LiudmylaKulinchenko: It's ok! Look in the checklist from the task. There are a few examples about braces location. Now you have "Bad example", like the next:
```
my_dict = {"greeting": "Good morning, have a nice day!", 
	   "answer": "Good morning, thanks!"
}
```
So your braces should be located in **one** of two options: open and start with the text **or** have a line break between the text. You have both options now

app/main.py
```diff
@@ -1,13 +1,19 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
```
LiudmylaKulinchenko, line 2: Don't forget that your braces should be located in **one** of two options: open and start with the text or have a line break between the text. You have **both** options here and in the next function 🐞

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! ☀

PR #619, opened 2023-05-24
My solution

kovaliskoveronika:


Oleg-s-s (changes requested):
You are duplicating your code several times, try to use already created functions in next ones.
tip: you can use 1st to handle error in 2nd and use 2nd to handle files in 3d

Oleg-s-s (changes requested):

Oleg-s-s (approved):
Well done :smile:

PR #620, opened 2023-05-24
Develop

Gvilion:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed"
```
Oleg-s-s, line 15: No need to check the len() here

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, linter_report[key])
+        for key in linter_report
```
Oleg-s-s, line 22: Use more descriptive variable name, not key

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if errors == [] else "failed"
```
LiudmylaKulinchenko, line 15: Empty list gives you false, so you don't have to check if errors equal to empty list

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🎈

PR #621, opened 2023-05-24
Solution

Tarasidze:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(err) for err in errors],
```
Oleg-s-s, line 13: It is better to use full names for variables if they are not so big

Oleg-s-s (changes requested):

Tarasidze (commented):
I've already done it. Check it, please

Oleg-s-s (approved):
Good job)

PR #623, opened 2023-05-24
Solution

Kasatello:


Oleg-s-s (changes requested):
You forgot to delete pass from functions, and be sure that you are passing flake8 test

PR #627, opened 2023-05-24
my_solution

Mr-Furman:


Oleg-s-s (changes requested):
Fix the code so it will pass a flake8 tests

PR #645, opened 2023-06-02
Solution linter formatter

MarkShapka:


app/main.py
```diff
@@ -1,13 +1,22 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if len([format_linter_error(error)
+                                   for error in errors]) == 0 else "failed"
```
Bohdan-Salamakha, line 16:
```suggestion
        "status": "passed" if not errors else "failed"
```

app/main.py
```diff
@@ -1,13 +1,22 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```

Bohdan-Salamakha (changes requested):
remove all __pycache__ folders from github

app/main.py
Bohdan-Salamakha: not fixed

Bohdan-Salamakha (changes requested):

olena-hul:
Good job!

PR #658, opened 2023-06-13
Solution

AndriiIshchenko:
flake8 returns error “Too long line…” on last “return” line
That’s why it is separated 
But I am not sure it is neсessary

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -17,6 +17,7 @@ def format_single_linter_file(file_path: str, errors: list) -> dict:
 
 
 def format_linter_report(linter_report: dict) -> list:
-    return \
-        [format_single_linter_file(path, err)
-         for path, err in linter_report.items()]
+    return [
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return \
+        [format_single_linter_file(path, err)
+         for path, err in linter_report.items()]
```
ArsenPidhoretskyi, line 22:
```suggestion
    return [
         format_single_linter_file(path, errors)
         for path, errors in linter_report.items()
     ]
```

ArsenPidhoretskyi (changes requested):
Can you try like this?

ArsenPidhoretskyi (approved):
Well done!

PR #676, opened 2023-06-23
Solution

a-kulaev:


app/main.py
```diff
@@ -1,13 +1,28 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error)
+            if errors != []
```
Polyakiv-Andrey, line 15: if and else are redundant here and should be omitted because if there are no errors, your list generator will return an empty list

Polyakiv-Andrey (changes requested):

app/main.py
a-kulaev: Thank You. I've changed the code

Polyakiv-Andrey (approved):
Good job!

PR #692, opened 2023-07-08
'Solution'

Lebwa1337:


app/main.py
```diff
@@ -1,13 +1,25 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```

Bohdan-Salamakha (changes requested):

app/main.py
Lebwa1337: How long i need to wait for yours respond on my changed solution?

Dimosphen1 (approved):
LGTM, well done!

PR #695, opened 2023-07-10
Test (py-linter-formatter)

baranovr:


app/main.py
```diff
@@ -1,13 +1,14 @@
...
-    # write your code here
-    pass
-
-
-def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
-
-
 def format_linter_report(linter_report: dict) -> list:
```
Polyakiv-Andrey, line 1: def format_linter_error(error: dict) -> dict: - handles 1 error
def format_single_linter_file(file_path: str, errors: list) -> dict: - calls the format_linter_error, passing arguments there, and adds new logic
def format_linter_report(linter_report: dict) -> list: - call format_single_linter_file and returns a list of the results of that function

Polyakiv-Andrey (changes requested):

PR #696, opened 2023-07-10
sk

ostboiko:


app/main.py
```diff
@@ -1,13 +1,18 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "columm": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"}
```
olena-hul, line 7:
```suggestion
        "source": "flake8"
    }
```

app/main.py
```diff
@@ -1,13 +1,19 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```

Bohdan-Salamakha (changes requested):
don't work on the master branch - it's a bad practice
https://github.com/mate-academy/py-task-guideline/blob/main/README.md#:~:text=git%20checkout%20%2Db%20develop

PR #697, opened 2023-07-11
Solution

valikdeb:
all test passed

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,20 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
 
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,20 @@
...
+        "errors": [format_linter_error(er) for er in errors],
+        "path": file_path,
+        "status": "passed" if errors == [] else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(k, v) for k, v in linter_report.items()]
```
Polyakiv-Andrey, line 20: k and v do not convey the meaning of the variable

Polyakiv-Andrey (changes requested):

app/main.py
valikdeb: done

pavlejviki (changes requested):
remove pycache from your git repository

PR #714, opened 2023-07-22
Implemented py-linter-formatter

kola686:


app/main.py
```diff
@@ -1,13 +1,26 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```
Bohdan-Salamakha: it's preferable to use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError

Bohdan-Salamakha (changes requested):

app/main.py
```diff
@@ -1,13 +1,26 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [
+        format_single_linter_file(report[0], report[1])
+        for report in linter_report.items()
```
```suggestion
        format_single_linter_file(key, value)
        for key, value in linter_report.items()
```

EdAlekseiev: Dict's method `items` returns 2 values, you can use unpacking for making your code more readable

EdAlekseiev (changes requested):

app/main.py
kola686: Got it, thanks!

EdAlekseiev (approved):
Good job

PR #723, opened 2023-07-25
Solution

VolodymyrSeniv:
Have a nice day)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,8 +1,11 @@
 def format_linter_error(error: dict) -> dict:
-    return {"line" : error["line_number"], 
-            "column" : error["column_number"], "message" : 
-            error["text"], "name": error["code"], 
-            "source" : "flake8"}
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,45 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```
Bohdan-Salamakha: it's preferable to use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
+            }
+            for error in errors
+        ],
```
Bohdan-Salamakha: use format_linter_error() function here

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8"
+                }
+                for error in file_errors
+            ],
+            "path": file_path,
+            "status": "failed" if file_errors else "passed"
+        }
```
Bohdan-Salamakha: use format_single_linter_file() function here

Bohdan-Salamakha (changes requested):

EdAlekseiev (approved):
Good job

PR #725, opened 2023-07-26
In the first function I return a dictionary with new keys. And values…

IProskurnytskyi:
… for these keys are taken from the input 'error' dictionary. The second function takes file_path and list of errors. Then it creates a dictionary with 'errors' key and value is a list of dictionaries  where each dictionary represents a single error but already formatted using 'format_linter_error'. 'status' is set to failed if 'errors' list is not empty. The third function processes each file in the linter_report dictionary by calling the format_single_linter_file function for each file.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if errors != [] else "passed"
```
Dimosphen1, line 15: You could limit this `if ... else` expression to `"failed" if errors else "passed"`. Remember that an empty list will already be faulty.

Dimosphen1 (approved):
LGTM, one minor comment is left to consider for the future.

app/main.py
IProskurnytskyi: Ok 
I did not think that I could do this instead.

PR #736, opened 2023-08-01
Implemented "format_linter_error", "format_single_linter_file" and "format_linter_report" functions!

raychw:
Implemented "format_linter_error", "format_single_linter_file" and "format_linter_report" functions!

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
```
pavlejviki: it's preferable to use get() method instead of parenthesis everywhere for getting value by key to avoid KeyError

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path, errors)
+        for file_path, errors in linter_report.items()
+    ]
```
pavlejviki: add a blank line at the end of the file

pavlejviki (changes requested):

pavlejviki (approved):
Good job!

PR #766, opened 2023-08-06
Solution

SerhiiBudko:


app/main.py
```diff
@@ -1,13 +1,36 @@
...
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if "errors" else "passed",
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
```
Polyakiv-Andrey: Use previous function instead rewrite code.

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1,13 +1,20 @@
...
+        "name": error["code"],
+        "source": "flake8",
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
```
Dimosphen1, line 13: You could consider rewriting this part with the use of `map()`

app/main.py
```diff
@@ -1,13 +1,20 @@
...
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(file_path, errors) for file_path, errors in linter_report.items()]
```
Dimosphen1: Flake8 check does not pass because this line is too long. Reformat the code.

Dimosphen1 (changes requested):
The flake8 check was not successful.

app/main.py
SerhiiBudko: Good day! All Your's recommendations were considered and already performed!

app/main.py
```diff
@@ -1,13 +1,25 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return list(
+        map(
+            lambda item: format_single_linter_file(item[0], item[1]),
```
Dimosphen1, line 22: You could use unpacking here, e.g. `lambda item: format_single_linter_file(*item)`

app/main.py
Dimosphen1: I suggested using `map()` only for this part of the code, e.g. `"errors": list(map(format_linter_error, errors))`. Both forms are acceptable.

app/main.py
Dimosphen1: I did not mean to use `map()` here, but your option is valid. However, your previous code looks more readable which is why prefer the previous form in the future.

Dimosphen1 (approved):
LGTM, the code is ligit. Several comments are left for future consideration.

PR #767, opened 2023-08-06
solution

ZaichenkoViktoriia:


Illia-Kononenko (approved):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
Y-Havryliv: key, value is not very informative, it's better to specify what is key and what is value

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file, linter_report[file])
+        for file in linter_report.keys()
```
Y-Havryliv, line 22: you don't need to specify `.keys()`. It iterates using keys by default

Y-Havryliv (approved):

PR #770, opened 2023-08-07
Implemented linter_formatter

osharlai:


app/main.py
```diff
@@ -1,13 +1,27 @@
...
+        "errors": [format_linter_error(one) for one in errors],
+        "path": file_path,
+        "status": "failed" if errors != [] else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
```
pavlejviki, line 20: use the existing format_single_linter_file function to avoid code duplication.

app/main.py
```diff
@@ -1,13 +1,27 @@
...
+    )
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(one) for one in errors],
+        "path": file_path,
+        "status": "failed" if errors != [] else "passed"
```
```suggestion
        "status": "failed" if errors else "passed"
```

pavlejviki: Empty lists evaluate to False in a boolean context. So you can just check the truthiness of the list to determine if it's empty or not. If errors is empty, it evaluates to False, otherwise True.

app/main.py
```diff
@@ -1,13 +1,27 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return dict(
+        line=error["line_number"],
+        column=error["column_number"],
+        message=error["text"],
+        name=error["code"],
+        source="flake8"
+    )
```
pavlejviki, line 9: You can simply write the dictionary and avoid using dict() method.

Using {}  is the more common way to define dictionaries.

pavlejviki (changes requested):

app/main.py
osharlai: I need to change 5 times "=" -> ":", then 5 times to put in "" each key of dict to fixed this simple comment)

app/main.py
osharlai: Done

osharlai: Ok, good advice

app/main.py
osharlai: Fixed code duplication

Polyakiv-Andrey (approved):
Good job!

PR #772, opened 2023-08-09
Task Errors Solution

weuis:


app/main.py
```diff
@@ -1,13 +1,30 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [
+        {
```
pavlejviki: use the already existing function   format_single_linter_file to avoid code duplication.

pavlejviki (changes requested):

PR #787, opened 2023-08-15
Solution

dmytrostepanenko:
Please, check my solution.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -10,13 +10,7 @@ def format_linter_error(error: dict) -> dict:
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
     return {
-        "errors": [
-            {
-                format_linter_error_key: format_linter_error_value
-                for format_linter_error_key, format_linter_error_value
-                in format_linter_error(error).items()
-            } for error in errors
-        ],
+        "errors": [format_linter_error(error) for error in errors],
         "path": file_path,
         "status": "failed" if len(errors) != 0 else "passed"
     }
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,27 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            {
+                key: value for key, value in format_linter_error(error).items()
```
vasylhnatiuk: use more accurate names then `key`, `value`

app/main.py
```diff
@@ -1,13 +1,27 @@
...
+        "status": "failed" if len(errors) != 0 else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path=key, errors=value)
+        for key, value in linter_report.items()
```
vasylhnatiuk: the same here

vasylhnatiuk (changes requested):

app/main.py
dmytrostepanenko: fixed

app/main.py
dmytrostepanenko: fixed

app/main.py
```diff
@@ -1,13 +1,29 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
```
vasylhnatiuk: you can use only` "errors": [format_linter_error(error) for error in errors]`

vasylhnatiuk (changes requested):

app/main.py
dmytrostepanenko: Thank you!

vasylhnatiuk (approved):

PR #803, opened 2023-08-21
Solution

narizhiy:


app/main.py
```diff
@@ -1,13 +1,26 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
```
olena-hul: use .get() instead of key access

olena-hul (changes requested):

app/main.py
narizhiy: I got you

narizhiy:
Done. I used .get() instead of key access

olena-hul (approved):

olena-hul:
Good job!

PR #813, opened 2023-08-24
Solution

asdadaversa:
please look at my solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,26 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,26 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if errors == [] else "failed"
```
Dimosphen1, line 17: You could rewrite this `if-else` statement without using `errors == []`; consider that an empty list is faulty.

Dimosphen1 (approved):

app/main.py
asdadaversa: "status": "failed" if errors else "passed"
?

PR #819, opened 2023-08-26
Solution

endernaga:


loralevitska (changes requested):
Please re-request review

app/main.py
```diff
@@ -1,13 +1,20 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
```
olena-hul, line 3: fix tests

olena-hul (changes requested):

app/main.py
endernaga: idk how to do it cause when I run test by my hand(from example) it works , but when I run automatic test it's dont work , and return TypeError
plus i checked type of argument and type function send str but not dict

EdAlekseiev (changes requested):
Fix tests, please

EdAlekseiev (approved):
Good job

PR #823, opened 2023-08-29
Solution

Familenko:


app/main.py
```diff
@@ -1,13 +1,19 @@
...
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if errors else "passed"}
```
LiudmylaKulinchenko: Fix indentation here, according to the best practices

app/main.py
```diff
@@ -1,13 +1,19 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(i, linter_report[i]) for i in linter_report
+    ]
```
LiudmylaKulinchenko: You also should use one indentation style throughout your code (open and start with the text or have a line break between the text)

LiudmylaKulinchenko (changes requested):

app/main.py
Familenko: thank you

LiudmylaKulinchenko (commented):
Please don't hide comments so that we can see them when reviewing your work next time.

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(i, linter_report[i]) for i in linter_report
```
LiudmylaKulinchenko: Use descriptive names instead of `i`. U can also unpack the items using `linter_report.items()` to increase readability

LiudmylaKulinchenko (changes requested):

app/main.py
Familenko: thanks!

wQuelS (approved):
GJ

PR #825, opened 2023-08-31
Solution

MatsalakO:


app/main.py
```diff
@@ -1,13 +1,24 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors" : [format_linter_error(error) for error in errors],
+        "path" : file_path,
+        "status" : "".join(["failed" if errors else "passed"][0])
```
Bohdan-Salamakha: u don't need to use join here
```suggestion
        "status" : "failed" if errors else "passed"
```

Bohdan-Salamakha (changes requested):

app/main.py
MatsalakO: I fixed this

pavlejviki (approved):
Well done!

PR #829, opened 2023-09-01
Solution

PyCore23:


app/main.py
```diff
@@ -1,13 +1,22 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {"line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"
+            }
```
pavlejviki, line 7:
```suggestion
    return {
        "line": error.get("line_number"),
        "column": error.get("column_number"),
        "message": error.get("text"),
        "name": error.get("code"),
        "source": "flake8"
    }
```

app/main.py
```diff
@@ -1,13 +1,22 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "failed" if errors else "passed"
+            }
```
pavlejviki, line 15:
```suggestion
return {
        "errors": [format_linter_error(error) for error in errors],
        "path": file_path,
        "status": "failed" if errors else "passed"
    }

```

app/main.py
```diff
@@ -1,13 +1,22 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path, error)
+        if error is not None else []
```
pavlejviki, line 20: if, else are redundant. This can lead to a list containing other lists and empty lists, which might not be the desired behavior.

pavlejviki (changes requested):

PR #835, opened 2023-09-07
Solution

Artem-user-19:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if not errors else "failed",
```
Dimosphen1, line 15: You could omit using `not` here and write `"failed" if errors else "passed"` instead.

Dimosphen1 (approved):
LGTM, well done! One minor comment is left for consideration.

PR #839, opened 2023-09-11
Solution

Damodar-kumar:


app/main.py
```diff
@@ -1,13 +1,45 @@
...
-    pass
+    return {
+        "errors": [
+            {
+                "line": error.get("line_number"),
+                "column": error.get("column_number"),
+                "message": error.get("text"),
+                "name": error.get("code"),
+                "source": "flake8",
+            }
```
ArsenPidhoretskyi: Reuse format_linter_error function

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+            for error in errors
+        ],
+        "path": "./source_code_2.py",
```
ArsenPidhoretskyi: No path can be any. Use file_path parameter

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+                    "column": source.get("column_number"),
+                    "message": source.get("text"),
+                    "name": source.get("code"),
+                    "source": "flake8",
+                }
+                for source in value
+            ],
+            "path": key,
+            "status": "passed" if value == [] else "failed",
+        }
```
ArsenPidhoretskyi: Reuse format_single_linter_file function instead of duplicate code

ArsenPidhoretskyi (changes requested):

Yurnerroo (approved):

PR #850, opened 2023-09-17
solution for format_linter_report

wataschi:
It's a rather confusing situation)

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,4 +1,3 @@
-
 def format_linter_error(error: dict) -> dict:
     return {
         "line": error.get("line_number"),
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,25 @@
...
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
 
 
+################################################################
+
```
Dimosphen1: Remove this part from your code, there is no need to have this separation here.

Dimosphen1 (changes requested):
One minor change was requested.

app/main.py
wataschi: I used it for convenience while writing code and forgot to delete it, sry

viktoria-rybenchuk (approved):
Well done!

PR #852, opened 2023-09-18
Solution

hbilous:
Solution amen

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,9 +1,9 @@
 def format_linter_error(error: dict) -> dict:
     return {
-        "line": error["line_number"],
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
```
olena-hul: Use .get() instead of key access

olena-hul (changes requested):

YukihiroSM (approved):
Good job!

PR #856, opened 2023-09-20
'Solution'

lylakivan:


app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8",
+    }
     pass
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
+    return {"errors": [format_linter_error(error) for error in errors],
```
viktoria-rybenchuk: add the first key from a new line

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+            "path": file_path,
+            "status": "failed" if errors else "passed"}
     pass
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
+    return [format_single_linter_file(key, value)
+            for key, value in linter_report.items()]
```
```suggestion
   return [
    format_single_linter_file(key, value)
    for key, value in linter_report.items()
    ]
```

viktoria-rybenchuk (changes requested):

app/main.py
viktoria-rybenchuk: it's a good practice to use descriptive variable names for both the key and value

lylakivan:
'Solution'

app/main.py
```diff
@@ -1,13 +1,27 @@
...
     pass
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
+    return [
+        format_single_linter_file(file_path, errors)
+        for file_path, errors in linter_report.items()
+    ]
     pass
```
Dimosphen1: Remove `pass` from your function body here and in other places, there is no need for it when the function is implemented.

Dimosphen1 (changes requested):
Several minor changes were requested.

lylakivan:
Solution

pavlejviki (approved):
Well done!

PR #862, opened 2023-09-25
Solution

D1st3f:
Added solution.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(i) for i in errors],
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path=i, errors=linter_report[i])
+        for i in linter_report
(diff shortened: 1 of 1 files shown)


Familenko:
short and correct!

Lyutillis:
Nice!

MiskoRuslan:
GJ!

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(i) for i in errors],
```
wQuelS: `i` is not a good variable name. Use more descriptive one, please

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed"
```
wQuelS: Can you try to simplify `len(errors) == 0` statement?
Consider a more "Pythonic" approach using some properties of containers ( such as lists, etc.), where an empty container is considered as `False`, and non-empty as `True`.

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path=i, errors=linter_report[i])
+        for i in linter_report
+    ]
```
wQuelS: 1. Same as above: bad variable name.
2. Iterating directly over a dictionary ( as you've done ) gives you its keys. But it seems like you need both keys and values. Consider using `.items()` method for such approach
3. keyword arguments are ok. They can increase readability and minimize some risks. However, consider using them throughout your whole codebase, because now it is not clear why here u've preferred keyword args, but the function above is being called with positional ones.

wQuelS (changes requested):
left some comments

wQuelS (approved):
GJ

PR #863, opened 2023-09-26
Solution

Lyutillis:


vshvanska (commented):
You should use list and dict comprehensions in this topic

YuliiaSaliuta (approved):

Shatrovskyi:
Good job! But In my opinion, in the second function, we don't necessarily use the ternary operator to check the status cause the function accepts only errors, and the status will be always "failed".

Shatrovskyi (approved):

app/main.py
```diff
@@ -1,13 +1,42 @@
...
+                {
+                    "line": error["line_number"],
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8",
+                } for error in errors
+            ],
+            "path": file,
+            "status": "failed" if errors else "passed",
```
LiudmylaKulinchenko: You are repeating yourself here. Just use the function that you create above. Don't duplicate your code.

app/main.py
```diff
@@ -1,13 +1,42 @@
...
-    # write your code here
-    pass
+    return {
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8",
```
LiudmylaKulinchenko: same here

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,13 +1,25 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error) for error in errors
+        ],
```
wQuelS: I guess this whole comprehension can be written in one line with brackets since it is not that long ?)

app/main.py
```diff
@@ -1,13 +1,25 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file, errors)
+        for file, errors in linter_report.items()
+    ]
```
wQuelS: `file` variable is a bit confusing because it is being passed to a function that takes `path` and `errors` as args)

wQuelS (changes requested):

oleksandr-pupchenko (approved):

skoval7 (approved):

LiudmylaKulinchenko (approved):
Good job 🎈☀

PR #867, opened 2023-09-26
Solution

MiskoRuslan:


vshvanska (approved):
Good job

app/main.py
```diff
@@ -1,13 +1,24 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(obj) for obj in errors],
```
wQuelS: isn't it just an `error` ?

app/main.py
```diff
@@ -1,13 +1,24 @@
...
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed"
```
wQuelS: `if len(errors) == 0` can be done in more "pythonic way" abusing property of collections ( like lists ) which is next: empty collection is considered as `False`, while non-empty as `True`

app/main.py
```diff
@@ -1,13 +1,24 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return \
+        [
+            format_single_linter_file(file_path, errors) for file_path,
+            errors in linter_report.items()
+        ]
```
wQuelS: something weird happened to the formatting here. please, fix it)

wQuelS (changes requested):

wQuelS (approved):
good

PR #868, opened 2023-09-26
Solution 

sakhaline:


mykolamateichuk (changes requested):
In `format_single_linter_file` you can remove the check for an empty `errors` list. It will not affect your function because if `errors` is empty your `for` cycle just won't run, and you'll still get an empty array. 

Good job, anyways!

app/main.py
```diff
@@ -1,13 +1,26 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error)
+            for error in errors
+        ] if errors else [],
```
wQuelS: this if statement is redundant, I agree

wQuelS (changes requested):

fedorov-s-od (approved):
Nice job with this improvement!

LiudmylaKulinchenko (approved):
Well done 🔥

PR #869, opened 2023-09-26
Solution: created 3 functions.

petrykivd:


OlehShumov (approved):

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(path, linter_report[path])
+            for path in linter_report.keys()]
```
wQuelS: So. if you've started to format like this:
image
please, follow this same style in other places

Also, your solution is correct, but what was behind the decision to iterate `.keys()` and then to access the corresponding values by each key in `linter_report[path]`?
I have a few points why iterating with unpacked key and value might be better:
- it is considered more "pythonic", when you need to access both keys and values. Also, it is more readable and straightforward.
- Your approach has similar efficiency for most use cases, but the approach with unpacking keys and values could be slightly more efficient because it avoids the extra dictionary lookup. ( While obtaining a value by key in a dictionary is O(1), skipping this extra lookup does have a minor efficiency benefit, saving resources by reducing the number of operations. )

what are your thoughts on this topic?

wQuelS (changes requested):

app/main.py
petrykivd: Thank you for review ;0 About my style, it's only mechanical mistake. I  used `.keys()` because it's first what i find when i Google " How to get values from dict". Right now i use `.items()`, I agree with you, my code will be faster by this.

wQuelS (approved):

PR #874, opened 2023-09-26
Solution

mykolamateichuk:


DenPrislipskyi (approved):
Nice solution!

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": ("passed" if not errors else "failed")
```
fsocie7y: I think parentheses are unnecessary here)

fsocie7y (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "path": file_path,
+        "status": "passed" if not errors else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path=file_path, errors=errors)
```
Tituha, line 21: I guess "=file_path" and "=errors" are also unnecessary

Tituha (approved):

app/main.py
wQuelS: > I guess "=file_path" and "=errors" are also unnecessary

it depends. Overall keyword args are a safer way to pass something to function, for example as they allow you to avoid some potential risks. But if you use such args, consider using them throughout your codebase to remain consistent

wQuelS (approved):
GJ

PR #877, opened 2023-09-27
Solution

Wonsky1:


katyakoziar (approved):

DenPrislipskyi (approved):
Nice solution!

app/main.py
```diff
@@ -1,13 +1,26 @@
...
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(
+            file_path=file_path,
+            errors=single_linter_file
+        )
+        for file_path, single_linter_file in linter_report.items()
+    ]
```
wQuelS: keyword args are okay, but I have concerns about a variable name that is passed to `format_single_linter_file` as an error. It is a bit confusing.

wQuelS (changes requested):

wQuelS (approved):

PR #878, opened 2023-09-27
Solution

Shatrovskyi:


RabenkoYevhenii (approved):
It`s seems to be a good solution

app/main.py
```diff
@@ -1,13 +1,33 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error)
+            for error in errors
+        ],
```
wQuelS: seems like this whole comprehension fits in one line including brackets, doesn't it?

app/main.py
```diff
@@ -1,13 +1,33 @@
...
-    pass
+    return [
+        {
+            "errors": [
+                format_linter_error(error)
+                for error in path_inside
+            ],
+            "path": path,
+            "status": "passed" if not path_inside else "failed"
+        }
```
wQuelS: code in this function is too complicated and can be simplified by using already-implemented functions

wQuelS (changes requested):

app/main.py
Shatrovskyi: Yep. Now It looks better.

app/main.py
Shatrovskyi: I made some improvements.

wQuelS (approved):
Nice

PR #880, opened 2023-09-27
Develop

roffi37:


Sel-Fisher (approved):
very beautifull, s kaifom👍

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "passed" if not errors else "failed",
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(fp, err)
+        for fp, err in linter_report.items()
```
wQuelS: those variable names are bad and not self-descriptive at all)

If you were trying to avoid writing `file_path` because of its length - you may consider define it as `path`,
`err` variable needs just 2 letters to become relevant ( from the PEP8 perspective )

wQuelS (changes requested):

wQuelS (approved):
Nicely done!

PR #886, opened 2023-09-28
Solution

Kateryna-Bordonos17:


RVChornyy (approved):
Nice!

wQuelS (changes requested):
изображение
Your formatting style in the first function differs from style in two other functions. Please, try to be consistent with it and provide a fix

app/main.py
```diff
@@ -1,13 +1,24 @@
...
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, errors)
```
Andrii-Silenko: Maybe "key" is not the best name of varuable. I'd prefer "file_path" or "path_of_file

app/main.py
```diff
@@ -1,13 +1,24 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(errors[i])
+                   for i in range(len(errors))],
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
```
Andrii-Silenko, line 16: It's ok, but you can rewrite this condition as:
```suggestion
        "status": "failed" if len(errors) > 0 else "passed"
```

Andrii-Silenko: You can rewrite this condition as:
```suggestion
        "status": "failed" if errors else "passed"
```

Andrii-Silenko (approved):

RVChornyy (approved):

app/main.py
```diff
@@ -1,13 +1,24 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(errors[i])
```
fsocie7y, line 13: in cases like this u can use `for element in elements: ...` without indexes

fsocie7y (approved):

PR #893, opened 2023-09-30
'Solution'

Olala2104:


app/main.py
```diff
@@ -1,13 +1,22 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    formatted_error = {
```
ArsenPidhoretskyi, line 2: Use just return 
```suggestion
   return {
```

ArsenPidhoretskyi: Fix in all places

ArsenPidhoretskyi (changes requested):
Not all data filed due to task requirements

PR #903, opened 2023-10-04
Solution

KwintDimon:


app/main.py
```diff
@@ -1,13 +1,26 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [format_linter_error(error) for error in errors],
```
Dimosphen1, line 15: You could use the `map` function here as well instead of list comprehension, e.g. 
```suggestion
        "errors": list(map(format_linter_error, errors)),
```

Dimosphen1 (approved):

PR #906, opened 2023-10-07
Solution

CrazyHilly:


app/main.py
```diff
@@ -1,13 +1,47 @@
...
+        "errors":
+            [
+                {
+                    "line": error["line_number"],
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8",
+                }
+                for error in errors
```
YukihiroSM: use format_linter_error instead of duplicating code.

app/main.py
```diff
@@ -1,13 +1,47 @@
...
+                        "message": error["text"],
+                        "name": error["code"],
+                        "source": "flake8",
+                    }
+                    for error in errors
+                ],
+            "path": file_name,
+            "status": "passed" if not errors else "failed"
+        }
+        for file_name, errors in linter_report.items()
```
YukihiroSM: Use format_single_linter_file

YukihiroSM (changes requested):

app/main.py
CrazyHilly: that's right, thanks! I thought about it at some point but then forgot :)

app/main.py
```diff
@@ -1,13 +1,24 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_name, errors)
+        for file_name, errors
+        in linter_report.items()
```
```suggestion
        for file_name, errors in linter_report.items()
```

ArsenPidhoretskyi (approved):
Good job, but check formating

PR #909, opened 2023-10-09
'Solution'

vpanchenkoSPS:


app/main.py
```diff
@@ -1,13 +1,25 @@
...
+        "status": "failed" if len(errors) > 0 else "passed"
+    }
     pass
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
+    return [
+        format_single_linter_file(file_path, errors)
+        for file_path, errors in linter_report.items()]
```
LiudmylaKulinchenko: Fix fomatting here

app/main.py
```diff
@@ -1,13 +1,25 @@
...
+    }
     pass
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
```
LiudmylaKulinchenko: You can check using `if errors` instead of length

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,13 +1,26 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
     pass
```
LiudmylaKulinchenko: You don't need `pass` if you have a docstring or function implementation

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Nice 🌺🌼🎈

PR #911, opened 2023-10-10
Solution

Taran21V:


app/main.py
```diff
@@ -1,13 +1,30 @@
...
+        "errors": [format_linter_error(error)
+                   for error in errors
+                   if error["filename"] == file_path
+                   ],
+        "path": file_path,
+        "status":
+        "failed" if [format_linter_error(error)
+                     for error in errors
+                     if error["filename"] == file_path
+                     ] else "passed"
```
YukihiroSM: What about setting this list to some variable instead of using 2 for loops?

YukihiroSM (changes requested):

EdAlekseiev (approved):
Good job

PR #920, opened 2023-10-15
Solution

fposlushnyi:
5.3 List and Dict Comprehensions

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,25 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    return {"line" if key == "line_number" else
-            "column" if key == "column_number" else
-            "message" if key == "text" else
-            "name" if key == "code" else
-            "source": "flake8" if key == "physical_line" else
-            val for key, val in (error.items())}
+    return {
+        {
+            "code": "name",
+            "text": "message",
+            "line_number": "line",
+            "column_number": "column",
+            "filename": "source"
+        }[key]: ("flake8" if key == "filename" else val)
+        for (key, val) in error.items()
+        if key != "physical_line"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
     return {"errors": [format_linter_error(error) for error in errors],
             "path": file_path,
-            "status": "passed" if len(errors) == 0 else "failed"}
+            "status": "passed" if errors == [] else "failed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    return [
-        {
-            "errors": linter_report[key],
-            "path": key,
-            "status": "passed" if (len(linter_report[key]) == 0) else "failed"
-        } if item == [] else
-        format_single_linter_file(key, linter_report[key])
-        for (key, item) in linter_report.items()
-    ]
+    return [format_single_linter_file(key, linter_report[key])
+            for (key, item) in linter_report.items()
+            ]
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,25 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {"line" if key == "line_number" else
+            "column" if key == "column_number" else
+            "message" if key == "text" else
+            "name" if key == "code" else
+            "source": "flake8" if key == "physical_line" else
+            val for key, val in (error.items())}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "passed" if len(errors) == 0 else "failed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": linter_report[key],
+            "path": key,
+            "status": "passed" if (len(linter_report[key]) == 0) else "failed"
+        } if item == [] else
+        format_single_linter_file(key, linter_report[key])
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,25 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {"line" if key == "line_number" else
+            "column" if key == "column_number" else
+            "message" if key == "text" else
+            "name" if key == "code" else
+            "source": "flake8" if key == "physical_line" else
+            val for key, val in (error.items())}
```
Dimosphen1: Reformat your method so that you will have a hardcoded structure with key and value without dict comprehension and `if else` statements here, it makes reading this method difficult.

app/main.py
```diff
@@ -1,13 +1,25 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "passed" if len(errors) == 0 else "failed"}
```
Dimosphen1: You could reformat this line considering that the empty list is faulty. You will not need to check the length of the list then.

app/main.py
```diff
@@ -1,13 +1,25 @@
...
-    # write your code here
-    pass
+    return [
+        {
+            "errors": linter_report[key],
+            "path": key,
+            "status": "passed" if (len(linter_report[key]) == 0) else "failed"
+        } if item == [] else
+        format_single_linter_file(key, linter_report[key])
+        for (key, item) in linter_report.items()
```
Dimosphen1: Reformat this method to utilize the `format_single_linter_file` method and avoid code duplication.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        for (key, val) in error.items()
+        if key != "physical_line"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "passed" if errors == [] else "failed"}
```
viktoria-rybenchuk: don't need compare error with [].
You can just check `if errors`

app/main.py
```diff
@@ -1,13 +1,23 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(key, linter_report[key])
+            for (key, item) in linter_report.items()
+            ]
```
viktoria-rybenchuk: Why do you need parentheses?
Use more descriptive names for 'key' and 'value'

```suggestion
        return [
        format_single_linter_file(key, linter_report[key])
        for (key, item) in linter_report.items()
    ]
```

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+    return {
+        {
+            "code": "name",
+            "text": "message",
+            "line_number": "line",
+            "column_number": "column",
+            "filename": "source"
+        }[signature]: ("flake8" if signature == "filename" else sense)
+        for signature, sense in error.items()
+        if signature != "physical_line"}
```
viktoria-rybenchuk: you should extract specific pieces of information from the input dictionary by key

app/main.py
```diff
@@ -1,13 +1,22 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "failed" if errors else "passed"}
```
viktoria-rybenchuk: While creating a dictionary — write key-value pairs in a single row. The curly braces must be located in one of two options: open and start with the text or have a line break between the text.

viktoria-rybenchuk (changes requested):
Read checklist.

app/main.py
```diff
@@ -1,13 +1,21 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        ["name", "source", "line", "column", "message"][i]:
+        ("flake8" if key == "filename" else error[key])
+        for i, key in enumerate(error) if key != "physical_line"
+    }
```
pavlejviki, line 8: you should hardcode the dictionary the same way as you did in def format_single_linter_file

app/main.py
```diff
@@ -1,13 +1,21 @@
...
```
pavlejviki, line 9:
```suggestion
 return {
        "line": error["line_number"],
         ...
    }
```

pavlejviki: and the same for all required keys

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "name": error["code"],
+        "source": "flake8",
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"]
```
Dimosphen1, line 7: Your `source` key should be placed at the last position as mentioned in the task. Pay attention to key positioning in the future.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

app/main.py
fposlushnyi: ok

PR #923, opened 2023-10-17
Все зробив і через день запустив ту саму програму і всі тести пройшли. Магія

eLQeR:


app/main.py
```diff
@@ -1,13 +1,19 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "failed" if [format_linter_error(error) for error in errors] else "passed"
```
YukihiroSM:
```suggestion
            "status": "failed" if errors else "passed"
```

YukihiroSM (changes requested):
Fix flake8 as well. 
image


Also, please, use commit messages which reflect what have you done on the task, not just your comments about how are the tests.

eLQeR:
Відкриваю

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+            "status": "failed" if errors else "passed"
+            }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, error)
+        for key, error in linter_report.items()
```
ArsenPidhoretskyi: key is not good name, rename to file_path

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "failed" if errors else "passed"
+            }
```
```suggestion
    return {
        "errors": [format_linter_error(error) for error in errors],
        "path": file_path,
        "status": "failed" if errors else "passed"
     }
```

app/main.py
ArsenPidhoretskyi: error -> errors
Because into linter_report as value list of error

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #928, opened 2023-10-18
Solution

Artkom98:


app/main.py
```diff
@@ -1,13 +1,26 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": len(errors) > 0 and "failed" or "passed"
```
YukihiroSM: This line will always return just True. 
Read more about using ternary operators. 

```suggestion
        "status": "failed" if errors else "passed"
```

YukihiroSM (changes requested):

YukihiroSM (approved):
Looks good to me!

PR #932, opened 2023-10-22
Solution for py-linter-formatter

anastasiiakalini:


app/main.py
```diff
@@ -1,13 +1,19 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"], "column": error["column_number"],
+        "message": error["text"], "name": error["code"], "source": "flake8"
+    }
```
pavlejviki, line 9: for better readability write each key from new line

app/main.py
```diff
@@ -1,13 +1,19 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path, "status": "failed" if errors != [] else "passed"
```
pavlejviki: for better readability write each key from new line

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if errors != [] else "passed"
```
```suggestion
        "status": "failed" if errors else "passed"
```

pavlejviki: an empty list is considered False when evaluated in a Boolean context, while a non-empty list is considered True

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #937, opened 2023-10-24
Solution

Marinel444:


app/main.py
```diff
@@ -1,13 +1,42 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            {
+                "line": error["line_number"],
```
pavlejviki: use existing format_linter_error function to avoid code duplication

app/main.py
```diff
@@ -1,13 +1,42 @@
...
+        "status": "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": [
```
pavlejviki: use existing format_single_linter_file function to avoid code duplication

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path=file, errors=linter_report[file])
+        for file in linter_report
```
pavlejviki: use items() so you can iterate over the  key and value of the dictionary

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

pavlejviki (approved):
Well done!

PR #939, opened 2023-10-25
finally solution

GibbsVlad:
Please can you remind me, this string "write your code here" I should delete, as it was in previous tasks, or should keep on their place as in tasks on mate.academy? =)


app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "message": error["text"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors
+                       if file_path == error["filename"]],
```
EdAlekseiev: This if is redundant

```suggestion
    return {"errors": [format_linter_error(error) for error in errors],
```

EdAlekseiev (changes requested):

GibbsVlad:
@EdAlekseiev  Is it correct?

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+            "path": file_path,
+            "status": "failed" if errors else "passed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, errors) for key,
+        errors in linter_report.items()
```
```suggestion
        format_single_linter_file(key, errors) 
        for key, errors in linter_report.items()
```
key - is bad variable name. In this case rename to file_path

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "failed" if errors else "passed"}
```
```suggestion
    return {
        "errors": [format_linter_error(error) for error in errors],
        "path": file_path,
        "status": "failed" if errors else "passed"
    }
```

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):
Nicely done! Thanks for fixing things!

PR #948, opened 2023-10-30
Solution

gertsovenko:


app/main.py
```diff
@@ -1,13 +1,20 @@
...
+        "column": error.get("column_number"),
+        "message": error.get("text"),
+        "name": error.get("code"),
+        "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
```
Polyakiv-Andrey: After opening the parentheses, write each the arguments on a new line, including the first one, the closing bracket is also written on a new line

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if errors == [] else "failed"
```
```suggestion
        "status": "failed" if errors else "passed"
```

viktoria-rybenchuk: this approach is preferred for its simplicity, readability, and maintainability, making it a more efficient

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(path, linter_report[path])
+        for path in linter_report
```
pavlejviki: you can use items() to iterate over the key and its value at the same time. so you can replace linter_report[path] with its actual value

pavlejviki (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #953, opened 2023-11-01
Solution

AlisaSkl:


app/main.py
```diff
@@ -1,13 +1,19 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return ({"line" if key == "line_number"
+            else "column" if key == "column_number"
+            else "message" if key == "text"
+            else "name" if key == "code" else key: value
+            for key, value in error.items()
+            if key not in ("filename", "physical_line")}
+            | {"source": "flake8"})
```
Dimosphen1: Remove the `if else` conditions and the dict comprehension. Hardcode the structure of the dict that you return instead for better readability. There is no need to place it inside of `()`, return it directly.

app/main.py
```diff
@@ -1,13 +1,19 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return ({"errors": [format_linter_error(error) for error in errors]}
+            | {"path": file_path}
+            | {"status": "passed" if len(errors) == 0 else "failed"})
```
Dimosphen1: Do not join several dicts, create one dict with the required keys at once. Besides that, do not place the dict inside of the `()`, and do not forget that an empty list is faulty by default, so there is no need to check its length.

app/main.py
```diff
@@ -1,13 +1,19 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(key, value)
+            for key, value in linter_report.items()]
```
Dimosphen1: Use more descriptive variable names instead of `key` and `value`.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,13 +1,17 @@
...
+            "name": error["code"],
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "passed" if errors == [] else "failed"}
```
Dimosphen1: You do not need to evaluate an empty list, it will be faulty by default. Change your code as follows:
```suggestion
            "status": "failed" if errors else "passed"}
```

app/main.py
```diff
@@ -1,13 +1,17 @@
...
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
```
Dimosphen1: You could use an alternative syntaxis here as well, just as an option:
```suggestion
    return list(map(format_linter_error, errors)),
```

Dimosphen1 (changes requested):
Several minor changes were requested.

app/main.py
AlisaSkl: done

app/main.py
AlisaSkl: done

pavlejviki (approved):
Good job!

PR #955, opened 2023-11-02
Solution

rinotokun:


app/main.py
```diff
@@ -1,13 +1,27 @@
...
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": (
+            "passed"
+            if errors == [format_linter_error(error) for error in errors]
+            else "failed"
```
viktoria-rybenchuk: you don't need call here function `format_linter_error`. You can just check `if errors`

app/main.py
```diff
@@ -1,13 +1,27 @@
...
+        )
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
viktoria-rybenchuk: it's a good practice to use more descriptive variable names to improve code readability and maintainability instead of `key`  and `value`

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if errors == [] else "failed"
```
viktoria-rybenchuk: you don't need to check for an empty list specifically, but rather, you cab to check if the `errors` variable

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(error_info_key, error_info_value)
+        for error_info_key, error_info_value in linter_report.items()
```
```suggestion
        for path, error  in linter_report.items()
```

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
```
Dimosphen1, line 13: You could use an alternative approach here as well, e.g.
```suggestion
        "errors": list(map(format_linter_error, errors)),
```

Dimosphen1 (approved):

PR #958, opened 2023-11-06
Develop

kolya2731:


app/main.py
```diff
@@ -1,13 +1,70 @@
...
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return format_single_linter_file_2(file_path, errors)
```
pavlejviki: no need to create additional functions. use format_linter_error inside this functions and hardcode the dictionary that it is supposed to return. hint: use list comprehension to apply format_linter_error to each error inside the errors list

app/main.py
```diff
@@ -1,13 +1,70 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return format_linter_report_2(linter_report)
```
pavlejviki: use  format_single_linter_file inside this function. no need to create additional functions

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #961, opened 2023-11-08
Solution

Grotski:


app/main.py
```diff
@@ -1,13 +1,45 @@
...
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
+            }
+            for error in errors
+        ]
```
EdAlekseiev: You should use function `format_linter_error ` to avoid code duplication

```suggestion
        "errors": [
            format_linter_error(error)
            for error in errors
        ]
```

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "path": file_path,
+        "status": "failed",
```
EdAlekseiev: `status` has to be `failed` if `errors` is not empty. In the other case, it has to be `passed`.

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8"
+                }
+                for error in errors
+            ],
+            "path": file_path,
+            "status": "failed" if errors else "passed"
+        }
```
EdAlekseiev: You should use function format_single_linter_file to avoid code duplication

EdAlekseiev (changes requested):

app/main.py
Grotski: I fixed it

Grotski:
Hi @EdAlekseiev  sorry, 
u see my last commit?

EdAlekseiev (approved):
Good job

PR #976, opened 2023-11-17
Solution

MariiaSydoruk:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(err) for err in errors],
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed",
```
```suggestion
        "status": "passed" if not errors else "failed",
```

pavlejviki: If errors contains any items, it evaluates to True

pavlejviki (changes requested):

app/main.py
MariiaSydoruk: What evaluates to True?

PR #978, opened 2023-11-23
total

Binkovskyi:


app/main.py
```diff
@@ -1,13 +1,24 @@
...
+        "name": error.get("code"),
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(i) for i in errors],
```
ArsenPidhoretskyi, line 13: i is not good name

ArsenPidhoretskyi (changes requested):

app/main.py
Binkovskyi: i need to correct it? or it's just notice to future

PR #980, opened 2023-11-23
Solutin

Olevsky-Ivan:
i belive that it's all

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,3 @@
+# Default ignored files
+/shelf/
+/workspace.xml
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -1,0 +1,25 @@
+<component name="InspectionProjectProfileManager">
+  <profile version="1.0">
+    <option name="myName" value="Project Default" />
+    <inspection_tool class="PyPep8NamingInspection" enabled="true" level="WEAK WARNING" enabled_by_default="true">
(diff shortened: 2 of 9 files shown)


app/main.py
```diff
@@ -1,13 +1,20 @@
...
+            "path": file_path,
+            "status": "failed" if errors else "passed"
+            }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(key, value)
+            for key, value in linter_report.items()]
```
viktoria-rybenchuk: Consider using more descriptive variable names instead of `key` and `value`

app/main.py
```diff
@@ -1,13 +1,20 @@
...
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
```
viktoria-rybenchuk: place each key on a new line for better readability

app/main.py
```suggestion
    return [
    format_single_linter_file(key, value)
     for key, value in linter_report.items()
     ]
```

viktoria-rybenchuk: both approaches are valid from a syntactic perspective, the second one is generally preferred in many style guides and is considered a best practice for writing clean and readable  code.

viktoria-rybenchuk (changes requested):

PR #985, opened 2023-11-26
Solution

VladimirDolhyi:


ArsenPidhoretskyi (approved):
Good job, but read code so hard

VladimirDolhyi:
> Good job, but read code so hard

Ok. How I can do this easier for better reading?

PR #995, opened 2023-11-28
creation linter formatter functions

Vitalii-Haponiuk:


app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors]
+        if len(errors) > 0 else [],
```
viktoria-rybenchuk: no need make this check. The list comprehension will create an empty list, and no additional check is required

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
```
viktoria-rybenchuk: You can simple check `if errors`

viktoria-rybenchuk (changes requested):

app/main.py
Vitalii-Haponiuk: understood, made changes

app/main.py
Vitalii-Haponiuk: ok

Dimosphen1 (approved):
LGTM, well done!

PR #996, opened 2023-11-29
Solution

dissom:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "passed" if errors == [] else "failed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(key, value)
+            for key, value in linter_report.items()]
```
vsmutok: key, value bad variable  name in this case

app/main.py
```diff
@@ -1,13 +1,21 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "line": error["line_number"],
```
vsmutok:
```suggestion
        "line": error.get("line_number"),
```

vsmutok: better to use `get()`

vsmutok: Please  fixed  everywhere

app/main.py
vsmutok: use file_path, errors

vsmutok (changes requested):

app/main.py
vsmutok: better to use get()

vsmutok: Please fixed everywhere

vsmutok (changes requested):

vsmutok (commented):
Good Job 👍

vsmutok (approved):

PR #997, opened 2023-11-29
'Solution'

Vik-By:


app/main.py
```diff
@@ -1,13 +1,38 @@
...
+            "column": linter_report[file_error][key_error]["column_number"],
+            "message": linter_report[file_error][key_error]["text"],
+            "name": linter_report[file_error][key_error]["code"],
+            "source": "flake8"}
+            for key_error, key in enumerate(linter_report[file_error])],
+            "path": file_error,
+            "status": "passed"
+            if len(linter_report[file_error]) == 0 else "failed"}
+        for file_error in linter_report
+    ]
```
Dimosphen1, line 23: The aim of this task is to use encapsulation and follow DRY principle. Use previously created functions in the other ones to not duplicate the same code fragments.

app/main.py
```diff
@@ -1,13 +1,38 @@
...
+    return {
+        "errors": [{
+            "line": every["line_number"],
+            "column": every["column_number"],
+            "message": every["text"],
+            "name": every["code"],
+            "source": "flake8"
+        } for every in errors if every["filename"] == file_path],
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
```
Dimosphen1: There is no need to evaluate the length of the list, an empty list is faulty by default. Rewrite this part to encounter that.

Dimosphen1 (changes requested):

app/main.py
Vik-By: If i call this function in next function, i must have this status F or T.

Dimosphen1: There is no need to evaluate the length, it will not change the logic. Change the code to the following:
```suggestion
        "status": "failed" if errors else "passed"
```

app/main.py
```diff
@@ -1,13 +1,27 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error=every) for every in errors
+            if every["filename"] == file_path
+        ],
```
Dimosphen1: There is no need in the if condition, all errors will correspond to the `file_path` because you pass them to the function in this way. Simplify this code:
```suggestion
        "errors": [format_linter_error(error) for error in errors],
```

app/main.py
```diff
@@ -1,13 +1,27 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path=file_error,
+                                  errors=linter_report[file_error])
+        for file_error in linter_report
```
Dimosphen1: Use `linter_report.items()` in the iteration with the use of unpacking to retrieve the key and its value at once without the need to obtain value by key later.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #998, opened 2023-11-30
Solution

maxbidun:


app/main.py
```diff
@@ -1,13 +1,22 @@
...
+            "name": error["code"],
+            "source": "flake8"
+            }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {"errors": [format_linter_error(i) for i in errors],
```
ArsenPidhoretskyi: i is not good name, rename to error

app/main.py
```diff
@@ -1,13 +1,22 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [{"errors": linter_report[i], "path": i, "status": "passed"}
+            if linter_report[i] == []
+            else format_single_linter_file(i, linter_report[i])
+            for i in linter_report]
```
ArsenPidhoretskyi: So strange code, simplify it if statement is redundant here, just use format_single_linter_file

ArsenPidhoretskyi (changes requested):

app/main.py
maxbidun: Because of the operator, my status changes
or should i do it in format_single_linter_file

app/main.py
```diff
@@ -1,13 +1,20 @@
...
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path, "status": "failed" if errors else "passed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [format_single_linter_file(file_path, error)
+            for file_path, error in linter_report.items()]
```
ArsenPidhoretskyi, line 20: error -> errors

ArsenPidhoretskyi (approved):
Good job!

PR #999, opened 2023-12-01
Solution

liudmylaperchevska:


app/main.py
```diff
@@ -1,13 +1,46 @@
...
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
+            }
+            for error in errors
+        ],
```
viktoria-rybenchuk: You have already defined this logic in previous function. Use ` format_linter_error` to avoid duplication

app/main.py
```diff
@@ -1,13 +1,46 @@
...
+        "path": file_path,
+        "status": ("passed" if errors == [] else "failed")
```
viktoria-rybenchuk: No need to explicitly check if `errors == []` because the conditional expression automatically evaluates to False when the `errors` list is empty, and True otherwise. Use `if errors`

app/main.py
```diff
@@ -1,13 +1,46 @@
...
+                    "name": error["code"],
+                    "source": "flake8"
+                }
+                for error in error_value
+            ],
+            "path": error_key,
+            "status":
+                ("failed" if error_value != [] else "passed")
+        }
+        for error_key, error_value in linter_report.items()
```
viktoria-rybenchuk: same here. Use previous function. Read checklist

viktoria-rybenchuk (changes requested):
remove `for_me_test` from your repository

app/main.py
liudmylaperchevska: good hint!

pavlejviki (approved):
good job!

PR #1010, opened 2023-12-06
Solution

honelyq:


app/main.py
```diff
@@ -1,13 +1,21 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error.get("line_number", 0),
+        "column": error.get("column_number", 0),
+        "message": error.get("text", ""),
+        "name": error.get("code", ""),
```
Dimosphen1, line 6: It would be applicable to leave the default output of `None` in case the key is not present in the dict. But kudos to populating default values that could apply to the content!

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
```
Dimosphen1, line 13: You could use `map` here as well, e.g.:
```suggestion
        "errors": list(map(format_linter_error, errors)),
```

Dimosphen1 (approved):
LGTM, well done! Several comments were left for consideration.

PR #1013, opened 2023-12-08
08_12_2023_16_32

SerheiZhus:


app/main.py
```diff
@@ -1,13 +1,16 @@
...
+            "path": file_path, "status": "failed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [{"errors": [], "path": i, "status": "passed"}
+            if not linter_report[i]
+            else format_single_linter_file(i, linter_report[i])
+            for i in linter_report]
```
vasylhnatiuk: i is a bad name

app/main.py
```diff
@@ -1,13 +1,16 @@
...
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(i) for i in errors],
+            "path": file_path, "status": "failed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [{"errors": [], "path": i, "status": "passed"}
```
vasylhnatiuk:
```suggestion
    return [
        format_single_linter_file(path, errors)
        for path, errors in linter_report.items()
    ]
```

vasylhnatiuk: it looks better

vasylhnatiuk (changes requested):

PR #1024, opened 2023-12-11
Solution

kabust:


app/main.py
```diff
@@ -1,13 +1,22 @@
+error_vars = {"line": "line_number",
+              "column": "column_number",
+              "message": "text",
+              "name": "code",
+              "source": "flake8"}
+
```
Serhii-Leonenko: it's better to define it directly in comprehension

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+            "path": file_path,
+            "status": "failed" if errors else "passed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(file_path=path,
+                                      errors=linter_report[path])
+            for path in linter_report.keys()]
```
Serhii-Leonenko: image

as for me formating like this has better readability, but you have to use the same style in all project

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,22 +1,24 @@
...
-    return {k: (error[v] if v in error.keys() else v)
-            for k, v in error_vars.items()}
+    return {
+        k: (error[v] if v in error.keys() else v)
+        for k, v in {"line": "line_number",
+                     "column": "column_number",
+                     "message": "text",
+                     "name": "code",
+                     "source": "flake8"}.items()
+    }
```
Serhii-Leonenko: you don't need to use comprehension here and iterate through the error dict, just return the correctly formatted dict. 
Also, it's obligatory to check your code with the checklist, here you violate it because you have bad variable names. But in this case, you don't need them at all, just return the dict

`return {
    "line": error["line_number"],
    ...
    }`

Serhii-Leonenko (changes requested):

ArturPoltser (approved):

Serhii-Leonenko (approved):
Well done!

PR #1026, opened 2023-12-13
Develop

M4Xpy:


app/main.py
```diff
@@ -1,13 +1,30 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    # """ formats all errors for a particular file
+    # and adds the path key — path to the file,
+    # and the status key — "failed" if there are errors,
+    # "passed" if there are no errors:
+    # """
```
YukihiroSM: if you are using docstrings it's not needed to add them as comments using #
```suggestion
    """ formats all errors for a particular file
    and adds the path key — path to the file,
    and the status key — "failed" if there are errors,
    "passed" if there are no errors:
    """
```

YukihiroSM: https://www.programiz.com/python-programming/docstrings

YukihiroSM (changes requested):
The solution is good! I like your comments for functios. The only one question is about docstrings :)

app/main.py
M4Xpy: i know it ,
but  normal docstring dont  permit your  test ,
and i sended report about  it  already.

So , i must delete docstrings at all ?

EdAlekseiev (approved):
Good job

PR #1029, opened 2023-12-14
Solution

mykytaso:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "passed" if not errors else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(errors=value, file_path=key)
+        for key, value in linter_report.items()
```
vasylhnatiuk: use more accurate names instead of key, value

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #1030, opened 2023-12-14
Solution

rzaliznyi:


rzaliznyi:
There are so many solutions because my flake8 in vs code don't want to load, and i had to do commit and then check if tests at git hub have passed succesful

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [format_linter_error(error) for error in errors],
```
Dimosphen1, line 15: You could use the following approach as well:
```suggestion
        "errors": list(map(format_linter_error, errors)),
```

Dimosphen1 (approved):

PR #1039, opened 2023-12-18
'Solution'

Maria145:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "failed" if len(errors) != 0 else "passed"}
```
viktoria-rybenchuk: use one code style. Place each key-value on a new line as you have made in previous function

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if len(errors) != 0 else "passed"
```
Dimosphen1, line 15: You could simplify this condition by assuming that non empty list is truthy by its boolean type  
```suggestion
        "status": "failed" if errors else "passed"
```
Prefer this approach in the future.

Dimosphen1 (approved):

PR #1059, opened 2024-01-02
Solution

SemenetsSergii:


app/main.py
```diff
@@ -1,13 +1,27 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(
+            file_path=path, errors=linter_report.get(path)
+        )
+        for path in linter_report
```
Dimosphen1: Perform an iteration on `linter_report.items()` with unpacking to avoid using `get` to obtain a value from dict.

app/main.py
```diff
@@ -1,13 +1,27 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error) for error in errors
+        ] if errors else [],
```
Dimosphen1: There is no need for this if condition, the list comprehension would create an empty list anyway even if `errors` list is empty.

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #1061, opened 2024-01-02
solution

marta-pylypchuk:


VladKyianytsia (approved):
Correct!

app/main.py
```diff
@@ -1,13 +1,26 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(
+            file_path=path, errors=linter_report.get(path)
+        ) for path in linter_report
+    ]
```
NEkropo1: Using linter_report.get(path) to access the dictionary value is less efficient than directly iterating over key-value pairs so consider iterating on dict.items()

NEkropo1 (changes requested):
Consider using `dict.items()`

NEkropo1 (approved):
Great!

PR #1063, opened 2024-01-03
Solution

VladKyianytsia:


juliastetsko (approved):
Good job!

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+            }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path, errors)
+        for file_path, errors in linter_report.items()
+    ]
```
NEkropo1, line 23: You should follow same style on whole project, including brackets and return statements/

NEkropo1 (changes requested):
You should keep same styling

app/main.py
VladKyianytsia: Fixed

NEkropo1 (approved):
Good job!

PR #1082, opened 2024-01-05
Solution

isikach:


Serejka21 (commented):
Need to fix flake8 E302 and W391 error

OleksandrKorolenko (approved):
looks good

app/main.py
```diff
@@ -1,13 +1,29 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(err) for err in errors],
```
Y-Havryliv:
```suggestion
        "errors": [format_linter_error(error) for error in errors],
```
`error` would be better

app/main.py
```diff
@@ -1,13 +1,29 @@
...
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": format_single_linter_file(path, err)["errors"],
+            "path": path,
+            "status": format_single_linter_file(path, err)["status"]
+        } for path, err in linter_report.items()
```
Y-Havryliv:
```suggestion
        } for path, errors in linter_report.items()
```

app/main.py
```diff
@@ -1,13 +1,29 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": format_single_linter_file(path, err)["errors"],
```
Y-Havryliv: you can use `format_single_linter_file` to format a single file, e.g.
`format_single_linter_file(path, errors)`,
and it will return the needed data

Y-Havryliv (changes requested):

Y-Havryliv (changes requested):
.

app/main.py
isikach: good idea, thanks)

NEkropo1 (approved):
Good job!

PR #1083, opened 2024-01-05
Solution

kstorozhenko:


app/main.py
```diff
@@ -1,13 +1,27 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if "errors" else "passed"
```
tskopenko:
```suggestion
        "status": "failed" if errors else "passed"
```
but u don't use it in the last part, so it don't touch the `pytest`
btw, at the last `def` you paste it correct

tskopenko (approved):
All code done according to the rules

Serejka21 (approved):

app/main.py
Serhii-Leonenko: Fix this

app/main.py
```diff
@@ -1,13 +1,27 @@
...
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": [format_linter_error(error) for error in errors],
+            "path": path,
+            "status": "failed" if errors else "passed"
+        }
```
Serhii-Leonenko: DRY, you have the same code in the previous func

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Well done!

PR #1093, opened 2024-01-08
changes

IlliaBohomol:
changes


viktoria-rybenchuk (changes requested):
All tests should passed. Fix flake8

IlliaBohomol:
all test passed in pycharm, problem with node(must be 16 but i have 12) how i can fix it?

PR #1100, opened 2024-01-11
Solution

fatamorganaaa:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "passed" if errors == [] else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
ArsenPidhoretskyi: key, value is not good names for this context

ArsenPidhoretskyi (changes requested):

app/main.py
fatamorganaaa: fixed

ArsenPidhoretskyi (approved):
Well done!

PR #1111, opened 2024-01-13
'Solution'

Sonemon:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error)
+            for error in errors if error["filename"] == file_path],
```
EdAlekseiev: This check is redundant, because function get `errors` related to `file_path`
```suggestion
            for error in errors],
```

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+            "path": file_path,
+            "status": "failed"}
```
EdAlekseiev: `status` should be "passed" if `errors` is empty.
```suggestion
            "status": "failed" if errors else "passed"}
```

app/main.py
```diff
@@ -1,13 +1,21 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [{"errors":
+             format_single_linter_file(source, errors).get("errors", []),
+             "path": source,
+             "status": "failed" if len(errors) else "passed"}
+            for source, errors in linter_report.items()]
```
EdAlekseiev: You can simplify this method
```suggestion
    return [ format_single_linter_file(file_path, errors)
        for file_path, errors in linter_report.items()]
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #1113, opened 2024-01-14
implement format_linter_error, format_single_linter_file, format_linter_report functions

Darya-Savchenko:


NazarSheremet (approved):
Good job. The solution is correct. However, in the future, the title of the Pull Request should be `Solution`. You can add information about the changes in the description, but it's not necessary

PR #1123, opened 2024-01-18
Solution

kapitoshk4:


vasylhnatiuk (changes requested):
fix tests

app/main.py
```diff
@@ -1,13 +1,27 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(path, file)
+        for path, file in linter_report.items()
```
ArsenPidhoretskyi: is second argument is really file?

ArsenPidhoretskyi (changes requested):

app/main.py
kapitoshk4: > is second argument is really file?

no, I'll change that.

NikOneZ1 (approved):
Well done!

PR #1129, opened 2024-01-19
Implement py-linter-formatter

EugeneProtsenko:


app/main.py
```diff
@@ -1,13 +1,44 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors":
+        [
```
vasylhnatiuk: DRY, use format_linter_error here

app/main.py
```diff
@@ -1,13 +1,44 @@
...
+        ],
+        "path": file_path, "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
```
vasylhnatiuk: reuse format_single_linter_file here instead of duplicating

vasylhnatiuk (changes requested):

app/main.py
EugeneProtsenko: I've already fix task, do I need to create new puul request or I can just make a commits?

YukihiroSM (approved):
Nicely done!

PR #1137, opened 2024-01-21
Solution

Romanskyy:


app/main.py
```diff
@@ -1,13 +1,30 @@
...
-    pass
+    return [
+        {
+            "errors": [
+                format_linter_error(error)
+                for error in errors if len(errors) > 0
+            ],
+            "path": file_path,
+            "status": "failed" if len(errors) > 0 else "passed",
+        }
```
NazarSheremet: Use format_single_linter_file function which does the same

app/main.py
```diff
@@ -1,13 +1,30 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed",
```
NazarSheremet: Need to add a check of status, if the status is passed or failed

NazarSheremet (changes requested):
Need some changes

app/main.py
Romanskyy: done

app/main.py
Romanskyy: done

Romanskyy:
please, review my PR

NazarSheremet (approved):
Good job!

YuriiSmolii (approved):
LGTM

PR #1147, opened 2024-01-25
added solution

MAFIOZNIK1999:


app/main.py
```diff
@@ -1,13 +1,18 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {"line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"
+            }
```
LiudmylaKulinchenko: You should fix formatting here, so braces will be placed with text eather at the start and at the end

app/main.py
```diff
@@ -1,13 +1,18 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(i) for i in errors],
```
LiudmylaKulinchenko: Its better to use more descriptive names here and in 17-18 lines in loops

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌷

PR #1151, opened 2024-01-25
Solution

apmaz:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "path": file_path,
+        "status": "passed" if errors == [] else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(element, linter_report[element])
+            for element in linter_report]
```
pavlejviki: use items() to access key, value at the same time

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if errors == [] else "failed"
```
```suggestion
        "status": "passed" if not errors  else "failed"
```

pavlejviki: empty list evaluates to False

pavlejviki (changes requested):

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "path": file_path,
+        "status": "passed" if not errors else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(key, value)
+            for key, value in linter_report.items()]
```
ArsenPidhoretskyi: key, value is not clear names in this context, please rename them

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if not errors else "failed"
```
Dimosphen1: You could write this condition without using `not` as well 
```suggestion
        "status": "failed" if errors else "passed"
```

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(error, errors)
+            for error, errors in linter_report.items()]
```
Dimosphen1: The variable name is not correct - the key does not contain `error`, it contains the `file_path`. Rename the variable to accurately match its content.

Dimosphen1 (changes requested):
The change in the variable name was requested.

Dimosphen1 (approved):
LGTM, well done!

PR #1153, opened 2024-01-26
Develop

RomanVovchuk:
make this

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,9 +1,9 @@
 def format_linter_error(error: dict) -> dict:
     return {
-        "line": error.get("line_number"),
-        "column": error.get("column_number"),
-        "message": error.get("text"),
-        "name": error.get("code"),
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
         "source": "flake8"
     }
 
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,49 @@
...
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors":
+        [
+            {
+                "line": error["line_number"],
```
vasylhnatiuk: use format_linter_error here instead of duplicating

app/main.py
```diff
@@ -1,13 +1,49 @@
...
+
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [{
+        "errors":
```
vasylhnatiuk: use format_single_linter_file here instead of duplicating

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: do it

vasylhnatiuk (changes requested):

vsmutok (approved):
Well Done!

PR #1163, opened 2024-01-31
Solution - Fix empty list

Toltecos:
Fix empty list

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -12,7 +12,7 @@ def format_single_linter_file(file_path: str, errors: list) -> dict:
     return {
         "errors": [format_linter_error(error) for error in errors],
         "path": file_path,
-        "status": "failed" if len(errors) else "passed"
+        "status": "failed" if errors else "passed"
     }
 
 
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
```
Polyakiv-Andrey, line 2: use method get for getting value from dict

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if len(errors) else "passed"
```
```suggestion
        "status": "failed" if errors else "passed"
```

pavlejviki: empty list is equal to False in boolean context

pavlejviki (changes requested):

vasylhnatiuk (approved):

PR #1165, opened 2024-02-01
Solution

ggeorgiyy:


app/main.py
```diff
@@ -1,13 +1,33 @@
...
-    pass
+    return {
+        "errors": [
+            {
+                "line": error.get("line_number"),
+                "column": error.get("column_number"),
+                "message": error.get("text"),
+                "name": error.get("code"),
+                "source": "flake8"
+            }
```
NikOneZ1: Try to not repeat yourself and reuse format_linter_error instead

app/main.py
```diff
@@ -1,13 +1,33 @@
...
+            for error in errors
+        ],
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
```
NikOneZ1:
```suggestion
        "status": "failed" if errors else "passed"
```

NikOneZ1 (changes requested):

app/main.py
vasylhnatiuk: do it

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1169, opened 2024-02-02
created linter formatter

RomanHlodann:


app/main.py
```diff
@@ -1,13 +1,30 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(err)
+            for err in errors
+            if err["filename"] == file_path
```
pavlejviki: this if condition is redundant

app/main.py
```diff
@@ -1,13 +1,30 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(err)
+            for err in errors
```
```suggestion
            for error in errors
```

app/main.py
```diff
@@ -1,13 +1,30 @@
...
+            if err["filename"] == file_path
+        ],
+        "path": file_path,
+        "status":
+            "failed"
+            if [err for err in errors if err["filename"] == file_path]
```
pavlejviki: you just need to check if the list of errors is not empty, no need in the list comprehension

pavlejviki (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #1175, opened 2024-02-05
Solution

Sergunshot:


app/main.py
```diff
@@ -1,13 +1,16 @@
...
+            "path": file_path, "status": ("failed" if errors else "passed")}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [{"errors": [format_linter_error(fault)
+            for fault in linter_report[key]], "path": key,
+             "status": ("failed" if linter_report[key] else "passed")}
+            for (key, value) in linter_report.items()]
```
EdAlekseiev: You should use function `format_single_linter_file` to avoid code duplication

EdAlekseiev (changes requested):

app/main.py
Sergunshot: In function #3 input function format_sigle_linter_file

app/main.py
```diff
@@ -1,13 +1,14 @@
...
-    pass
+    return {"errors": [format_linter_error(mistake) for mistake in errors],
+            "path": file_path, "status": ("failed" if errors else "passed")}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(key, value)
+            for (key, value) in linter_report.items()]
```
vasylhnatiuk: key, value are bad names in this case

vasylhnatiuk (changes requested):

app/main.py
Sergunshot: Replace key == path, value = faults

vasylhnatiuk (approved):

PR #1187, opened 2024-02-11
Solution

MYevtushyk:


ArsenPidhoretskyi (approved):
.

app/main.py
```diff
@@ -1,13 +1,18 @@
...
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "passed" if errors == [] else "failed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(error, linter_report[error])
+            for error in linter_report]
```
ArsenPidhoretskyi, line 18: Use .items() instead of iteration only by key

ArsenPidhoretskyi (changes requested):

PR #1190, opened 2024-02-12
Solution

AlexTsikhun:


app/main.py
```diff
@@ -1,13 +1,29 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors":
+            [
+                format_linter_error(i) for i in errors
```
viktoria-rybenchuk: don't use single-letter variable

app/main.py
```diff
@@ -1,13 +1,29 @@
...
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(
+            file_path=k,
+            errors=v,
+        )
+        for k, v in linter_report.items()
```
viktoria-rybenchuk: the same here

app/main.py
```diff
@@ -1,13 +1,29 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors":
+            [
+                format_linter_error(i) for i in errors
+            ],
```
```suggestion
            [format_linter_error(i) for i in errors],
```

viktoria-rybenchuk (changes requested):

app/main.py
AlexTsikhun: I looked at the code and saw that code is bad) I'm going to correct other mistakes!

app/main.py
AlexTsikhun: > the same here

Should I create new commits for making different fixes or do it in one commit, what is good practice? (split commits by different logic?)

pavlejviki (approved):
Well done!

PR #1191, opened 2024-02-12
Solution

DmytroHlazyrin:


nicksetrakov (commented):
Use the previous functions you created, you are duplicating the code

nicksetrakov (changes requested):
Use the previous functions you created, you are duplicating the code

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if len(errors) else "passed"
```
YuriiKindrat: no need to use len here, it will work fine without it )

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Looks good! 🌸

PR #1223, opened 2024-02-21
Solution

klishchnazar:


YuriiSmolii (changes requested):
You should reuse functions inside of another functions.

app/main.py
```diff
@@ -1,13 +1,36 @@
...
+    return [
+        {
+            "errors": [],
+            "path": file_path,
+            "status": "passed"
+        }
+        if errors == []
+        else format_single_linter_file(file_path, errors)
+        for file_path, errors in linter_report.items()
+    ]
```
YuriiSmolii: No need to check for errors, you already have to check inside format_linter_error function
There should be something like that here:
```suggestion
    return [
        format_single_linter_file(file_path, errors) for file_path,
        errors in linter_report.items()
    ]
```

YuriiSmolii (changes requested):
Fix last function

YuriiSmolii (approved):
LGTM

PR #1227, opened 2024-02-21
Solution

o-zinevych:
I've checked my solution with the checklist. I decided not to put key-value pairs into one line because of worse readability. If there is a nicer way to solve it, please give me a tip.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -10,14 +10,9 @@ def format_linter_error(error: dict) -> dict:
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
     return {
-        key: (
-            [format_linter_error(error) for error in errors]
-            if key == "errors" else
-            file_path if key == "path" else
-            "failed" if errors else
-            "passed"
-        )
-        for key in ["errors", "path", "status"]
+        "errors": [format_linter_error(error) for error in errors],
+        "path": f"{file_path}",
+        "status": "failed" if errors else "passed"
     }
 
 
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,31 @@
...
+    return {
+        key: (
+            [format_linter_error(error) for error in errors]
+            if key == "errors" else
+            file_path if key == "path" else
+            "failed" if errors else
+            "passed"
+        )
+        for key in ["errors", "path", "status"]
+    }
```
```suggestion
    return {
        "errors": [format_linter_error(error) for error in errors],
        "path": f"{file_path}",
        "status": "failed" if errors else "passed"
    }
```

app/main.py
```diff
@@ -1,13 +1,31 @@
...
+    return {
+        key: (
+            error["line_number"] if key == "line" else
+            error["column_number"] if key == "column" else
+            error["text"] if key == "message" else
+            error["code"] if key == "name" else
+            "flake8"
+        )
+        for key in ["line", "column", "message", "name", "source"]
+    }
```
```suggestion
    return {
        "line": error["line_number"],
        "column": error["column_number"],
        "message": error["text"],
        "name": error["code"],
        "source": "flake8"
    }
```

EdAlekseiev (changes requested):

app/main.py
o-zinevych: Ooooh, I got it now. I wanted to do it like this initially, but I thought that I had to use some comprehension in each return

app/main.py
EdAlekseiev: Still not fixed

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #1232, opened 2024-02-24
Solution

D0mln0:


app/main.py
```diff
@@ -1,13 +1,42 @@
...
-    pass
+    return {
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
+            } for error in errors
```
ArsenPidhoretskyi: Reuse format_linter_error

app/main.py
```diff
@@ -1,13 +1,42 @@
...
+                    "line": error["line_number"],
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8"
+                } for error in linter_report[file_path]
+            ],
+            "path": file_path,
+            "status": "failed" if linter_report[file_path] else "passed"
+        } for file_path in linter_report
```
ArsenPidhoretskyi: Reuse format_single_linter_file

ArsenPidhoretskyi (changes requested):

app/main.py
D0mln0: i can`t understand how to fix err with "status"
it`s always "failed", i tried to put that if structure in "format_single_linter_file" but it doesn`t work, can you help with that?

YukihiroSM (approved):
Nicely done!

PR #1253, opened 2024-03-06
Solution

MishaCherpak:


app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "failed" if len(errors) > 0 else "passed"
```
vasylhnatiuk:
```suggestion
            "status": "failed" if errors else "passed"
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+            }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
vasylhnatiuk: use more accurate names instead of key, value

vasylhnatiuk: use more accurate names instead of key, value

vasylhnatiuk (changes requested):
use more accurate names instead of key, value

vasylhnatiuk (approved):

PR #1256, opened 2024-03-06
Solution

VladikVll:


app/main.py
```diff
@@ -1,13 +1,27 @@
...
+        "status": "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(report, linter_report[report])
+        for report in linter_report
```
Dimosphen1: Use `linter_report.items()` and perform unpacking to avoid retrieving value by key

app/main.py
```diff
@@ -1,13 +1,27 @@
...
-    pass
+    return {
+        "errors": [],
+        "path": file_path,
+        "status": "passed"
+    } if not errors else {
+        "errors": [format_linter_error(err) for err in errors],
+        "path": file_path,
+        "status": "failed"
+    }
```
Dimosphen1: It would be beneficial not to duplicate code and add value to only the `status` key using the if condition.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "passed" if not errors else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
vsmutok: `key`, `value` not clear variable  name

vsmutok (changes requested):

app/main.py
VladikVll: Дякую))

viktoria-rybenchuk (approved):
Good job!

PR #1265, opened 2024-03-10
Solution

panicua:
I hope it's fine cause all the way it felt like I'm doing something wrong :)

p.s. line 15-16 had to be splited/transfered in 2 cause flake8 hated that I used 82-83~ symbols instead of 79 even though it looks uglier than: "for key, value in {"errors": [format_linter_error(error) for error in errors]," imo

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -19,6 +19,6 @@ def format_single_linter_file(file_path: str, errors: list) -> dict:
 
 def format_linter_report(linter_report: dict) -> list:
     return [
-        format_single_linter_file(key, error)
-        for key, error in linter_report.items()
+        format_single_linter_file(file_path, error)
+        for file_path, error in linter_report.items()
     ]
--- a/app/main.py
+++ b/app/main.py
@@ -1,20 +1,15 @@
 def format_linter_error(error: dict) -> dict:
-    return {
-        key: value
-        for key, value in {"line": error["line_number"],
-                           "column": error["column_number"],
-                           "message": error["text"],
-                           "name": error["code"],
-                           "source": "flake8"}.items()}
+    return {"line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    return {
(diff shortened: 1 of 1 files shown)


panicua (commented):
flake8 didn't show me bad practice with braces, didn't see it at the time of first commit.

app/main.py
```diff
@@ -1,13 +1,24 @@
...
+                                      for error in errors],
+                           "path": file_path,
+                           "status": "failed" if errors else "passed"}.items()}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, error)
```
Boyok1337: use more informative names for variables
f.e. file_path, errors

Boyok1337 (changes requested):

app/main.py
panicua: I'll change *key* to *file_path*, but isn't it better to leave *error* instead of *errors* because it's a for cycle an we go through every item of that dict.

app/main.py
```diff
@@ -1,13 +1,24 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        key: value
+        for key, value in {"line": error["line_number"],
+                           "column": error["column_number"],
+                           "message": error["text"],
+                           "name": error["code"],
+                           "source": "flake8"}.items()}
```
LiudmylaKulinchenko: You iterate through a dict here to create a same dict
Remove the loop and leave just a dictionary

app/main.py
```diff
@@ -1,13 +1,24 @@
...
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        key: value
+        for key, value in {"errors": [format_linter_error(error)
+                                      for error in errors],
+                           "path": file_path,
+                           "status": "failed" if errors else "passed"}.items()}
```
LiudmylaKulinchenko: Same here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good! 💪

PR #1275, opened 2024-03-12
Solution

saywin:


app/main.py
```diff
@@ -1,13 +1,20 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path, "status": "failed"} \
+        if len(errors) != 0 \
+        else {"errors": [], "path": file_path, "status": "passed"}
```
```suggestion
    return {
        "errors": [format_linter_error(error) for error in errors],
        "path": file_path, 
        "status": "failed" if len(errors) != 0 else "passed"
     }
```
Simplify code and make it more readable

app/main.py
```diff
@@ -1,13 +1,20 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(key, value)
+            for key, value in linter_report.items()]
```
ArsenPidhoretskyi: key, value are not good names in this context

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #1278, opened 2024-03-12
Solution

mikhailyemets:


app/main.py
```diff
@@ -1,13 +1,26 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error.get("line_number", None),
```
oleksandra-shershen: If you are sure that the` line_number` key will always be present in the error dictionary, you can use 
`"line": error["line_number"]` instead of `"line": error.get("line_number", None)`. But it’s up to you.

oleksandra-shershen (approved):
Overall, the provided code appears well-structured and without repeated code. Good job!

app/main.py
YuriiKindrat: no need to write None as default arg, it will return None by default
fix this everywhere

YuriiKindrat (changes requested):

app/main.py
YuriiKindrat: no need to write None as default arg, it will return None by default
fix this everywhere

YuriiKindrat (changes requested):

YuriiKindrat (approved):
LGTM!

PR #1284, opened 2024-03-13
All function is done

Nyxan:


messedxup (approved):
Code style looks pretty good, code includes both allowed options for creating a dictionary and a list. Functions works in a proper way, according to the task. Good joob

LiudmylaKulinchenko (approved):
Good job 🌸

PR #1285, opened 2024-03-13
Solution 1

semermaks:


app/main.py
```diff
@@ -1,13 +1,31 @@
...
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
+    return [
+        {
+            "errors": [
+                format_linter_error(error) for error in errors
+            ],
+            "path": file_path,
+            "status": "failed" if errors else "passed"
```
dkotkod: this code duplicates the function format_single_linter_file, it can be replaced with a function, passing arguments to it, and thereby reducing the function code format_linter_report. And improve readability.

dkotkod (changes requested):
The general code passed tests  'pytest' and tests 'flake8'. But I would suggest some improvements. function 'format_linter_report' duplicates function 'format_linter_error', so it is better to call function 'format_linter_error' in function function 'format_linter_report', passing arguments there. then the code will become shorter and more readable

dkotkod (approved):
good job

app/main.py
```diff
@@ -1,13 +1,24 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
     pass
```
LiudmylaKulinchenko: `The pass statement is used as a placeholder for future code.
When the pass statement is executed, nothing happens, but you avoid getting an error when empty code is not allowed.` So, you don't need to keep those passes here, remove them

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🌸

PR #1289, opened 2024-03-13
Solution

Inna-Osavoliuk:


ihor-jpeg (changes requested):
test

Inna-Osavoliuk:
I added a new commit, but I couldn't update the task and send it for review.

PR #1293, opened 2024-03-13
Solution#

VitalyBashkiser:


app/main.py
yossosh, line 1: I think the solution is correct! ^^

yossosh (approved):

app/main.py
```diff
@@ -1,13 +1,18 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path, "status": "failed" if errors else "passed"}
```
YuriiKindrat, line 13: it is better to place one key/value per line for better readability

YuriiKindrat (changes requested):

app/main.py
YuriiKindrat: it is better to place one key/value per line for better readability

YuriiKindrat (changes requested):

app/main.py
YuriiKindrat: nothing changed (

YuriiKindrat (changes requested):

app/main.py
YuriiKindrat: still bad formatting here 12-13 line

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path, "status": "failed" if errors else "passed"
+    }
```
LiudmylaKulinchenko: The first comment for about to split each key:value by new line
you have 3 key/value pairs:
1)  "errors": [format_linter_error(error) for error in errors]
2)"path": file_path
3) "status": "failed" if errors else "passed"

And each of them should start from a new line. You need just to add a new line between 2) and 3) pair

```
{
        "errors": [format_linter_error(error) for error in errors],  # first key/value pair
        "path": file_path,  # second key/value pair
        "status": "failed" if errors else "passed"  # third key/value pair
    }
```
  Now each pair starts from a new line. U also have an example in the [checklist](https://github.com/mate-academy/py-linter-formatter/blob/master/checklist.md)
  If you still don't understand something, please, write us on the platform

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done 😌🌷

PR #1299, opened 2024-03-15
Solution

VladyslavKsh:


app/main.py
```diff
@@ -1,13 +1,28 @@
...
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error) for error in errors
+        ],
+        "path": file_path,
+        "status": "failed" if [
+            format_linter_error(error) for error in errors
+        ]
```
LiudmylaKulinchenko: In this function you get erros list as an argument. So you can just check if there are errors in this line

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,13 +1,30 @@
...
-    pass
+    return [
+        format_single_linter_file(file_path, errors)
+        for file_path, errors in linter_report.items()
+    ]
+
+
+# "status": "failed" if [
+#             format_linter_error(error) for error in errors
+#         ]
```
YuriiKindrat: remove unnecessary comments here

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Well done 💪

PR #1302, opened 2024-03-15
Solution

oleg-gj:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed",
```
viktoria-rybenchuk: there is no need to compare with 0, you can just check `if not errorrs`

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(path, errors) for path, errors in
+            linter_report.items()]
```
```suggestion
    return [
    format_single_linter_file(path, errors) for path, errors in linter_report.items()
    ]
```

viktoria-rybenchuk: fix formatting

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Well done!

app/main.py
Ja0ni: "failed" if errors else "passed"

PR #1330, opened 2024-04-03
Solution

juliazelena:


app/main.py
```diff
@@ -1,13 +1,82 @@
...
+        "name": error.get("code"),
+        "source": "flake8",
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors":
```
Polyakiv-Andrey, line 13: use the format_single_linter_file function here instead of repeating the code

app/main.py
```diff
@@ -1,13 +1,82 @@
...
+                    "line": error.get("line_number"),
+                    "column": error.get("column_number"),
+                    "message": error.get("text"),
+                    "name": error.get("code"),
+                    "source": "flake8",
+                }
+                for error in errors
+            ],
+        "path": file_path,
+        "status": "failed",
```
Polyakiv-Andrey: status should be depend from errors, if errors is absent status should be success

app/main.py
```diff
@@ -1,13 +1,82 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
```
Polyakiv-Andrey, line 23: use the format_single_linter_file function here and remove all unnecessary code

Polyakiv-Andrey (changes requested):

app/main.py
```diff
@@ -1,13 +1,35 @@
...
+        {
+            "errors": linter_report["./test_source_code_2.py"],
+            "path": "./test_source_code_2.py",
+            "status": "passed",
+        },
+        format_single_linter_file("./source_code_2.py",
```
vasylhnatiuk: use loop and don't use code like this `"./source_code_2.py"`, you should get data from `linter_report`

vasylhnatiuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #1336, opened 2024-04-07
Develop

AlexTarasov57:


app/main.py
```diff
@@ -1,13 +1,18 @@
...
+    return {
+        "errors": [format_linter_error(error=i) for i in errors],
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed"
+    }
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(file_path=key, errors=value) for key, value in linter_report.items()]
```
ArsenPidhoretskyi, line 18: key, value is not clear names in this context

app/main.py
```diff
@@ -1,13 +1,18 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
-
+    return {
+        "errors": [format_linter_error(error=i) for i in errors],
```
ArsenPidhoretskyi, line 12: i - is not good variable name in this context

ArsenPidhoretskyi (changes requested):

PR #1350, opened 2024-04-14
Solution

VadimkaLYZ:


app/main.py
```diff
@@ -1,13 +1,45 @@
...
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
+            }
+            for error in errors
+        ],
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
```
YukihiroSM: how about reusing format_linter_error here? 
For now you are duplicating the same code from the function above

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8"
+                }
+                for error in errors
+            ],
+            "path": file_path,
+            "status": "failed" if errors else "passed"
+        }
+        for file_path, errors in linter_report.items()
```
YukihiroSM: Same here. Reuse format_single_linter_file.

YukihiroSM (changes requested):

app/main.py
VadimkaLYZ: thanks, fixed it

vsmutok (approved):
Good Job!

PR #1355, opened 2024-04-19
Task completed, but pytest throws a weird error

CycleBreaker:
Task completed, but pytest throws a weird "string indices must be integers, not str" error, pointing at my format_linter_error function. It goes away when I actually use ints for indices, but cycling through keys to look for specific names instead of addressing the keys directly feels pretty dumb. Please advise.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -18,6 +18,6 @@ def format_single_linter_file(file_path: str, errors: list) -> dict:
 
 def format_linter_report(linter_report: dict) -> list:
     return [
-        [format_single_linter_file(path, error) for error in errors]
+        format_single_linter_file(path, errors)
         for path, errors in linter_report.items()
     ]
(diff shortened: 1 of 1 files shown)


vsmutok (changes requested):
tests failed!

CycleBreaker:
@vsmutok Yeah, but can you please comment on the issue I described above?

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        [format_single_linter_file(path, error) for error in errors]
```
```suggestion
        format_single_linter_file(path, errors)
```
So you should format the file

ArsenPidhoretskyi (changes requested):

app/main.py
CycleBreaker: Got it. Thanks!

ArsenPidhoretskyi (approved):
Well done!

PR #1365, opened 2024-05-01
Solution

BohdanKuzik:


app/main.py
```diff
@@ -1,13 +1,32 @@
...
+
+
+def format_single_linter_file(file_path: str, errors: list) -> dict:
+
+    return {
+        "errors": [
+            format_linter_error(error) for error in errors
+        ],
+        "path": file_path,
+        "status": "failed"
```
Oleg-s-s: You sure that status will always be failed?

app/main.py
```diff
@@ -1,13 +1,32 @@
...
+
+
+def format_linter_report(linter_report: dict) -> list:
+
+    return [
+        {
+            "errors": [],
+            "path": path,
+            "status": "passed",
+        } if not report else format_single_linter_file(path, report)
```
Oleg-s-s: You can optimize this cause your lines 27-29 almost same as 15-19

Oleg-s-s (changes requested):
Return all deleted files

BohdanKuzik:
> Return all deleted files



> Return all deleted files

I was on qna, and there I asked a question about the fact that I had files together with the solution at the same time, that is, all these files that I deleted, they were added again by accident, since there was a problem with git bash earlier, so it was recommended to delete them in the (cache) itself, so we did it together with the mentor and deleted the other 7 faults and he said that everything should be okay with the check.

app/main.py
```diff
@@ -1,13 +1,22 @@
...
-    pass
-
-
-def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+def format_linter_error(error: dict) -> dict:
+    return {
+        "line": error["line_number"], "column": error["column_number"],
+        "message": error["text"], "name": error["code"], "source": "flake8"
```
Oleg-s-s: Bad formatting, you should try to follow 1 of the structures:
1 - all in 1 line (not good here, will exceed the line len)
2 - 1 by 1 in a new line

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
+
+
+def format_linter_report(linter_report: dict) -> list:
+    return [
+        {"errors": [], "path": path, "status": "passed"}
+        if not report
+        else format_single_linter_file(path, report)
```
Oleg-s-s: You still are not reusing your code, now your `format_single_linter_file` is handling both cases if passed or not, this `if` is redundant

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
+
+def format_linter_report(linter_report: dict) -> list:
+    return [
+        format_single_linter_file(path, report)
+        for path, report in linter_report.items()
+    ]
```
Oleg-s-s: Be more patient, missed blank line(

Oleg-s-s (changes requested):

Oleg-s-s (approved):
:+1:

PR #1376, opened 2024-05-04
solution

Alex93Troy:


app/main.py
```diff
@@ -1,13 +1,41 @@
...
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
```
vasylhnatiuk: reuse format_linter_error here, don't duplicate your code

app/main.py
```diff
@@ -1,13 +1,41 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": [
+                {
+                    "line": error["line_number"],
```
vasylhnatiuk: reuse format_single_linter_file here

vasylhnatiuk (changes requested):

app/main.py
Alex93Troy: Okay, thank you. I edit code without duplicate

Alex93Troy: Task is done or somthing wrong ?

PR #1381, opened 2024-05-07
'Solution'

RuslanSamsonovych:


app/main.py
```diff
@@ -1,13 +1,17 @@
...
+            "path": file_path,
+            "status": "passed" if not errors else "failed", }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
EdAlekseiev: Instead of using generic variable names like `key` and `value`, consider choosing more descriptive names that clearly indicate their purpose. For example, you could use `file_path` and `errors`. These names make your code easier to read.

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #1384, opened 2024-05-09
Solution

AndrewwSSS:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if len(errors) else "passed"
```
Oleg-s-s: no need to check for `len()`, simple `if errors` will do the job

Oleg-s-s (changes requested):

wQuelS (approved):
LGTM

PR #1398, opened 2024-05-14
Solution

vladyslavgusakovskiy:


YuriiSmolii (changes requested):
You should reuse function `format_linter_error` inside `format_single_linter_file` function. Also reuse `format_single_linter_file` function inside `format_linter_report` function.

YuriiSmolii (approved):
LGTM

PR #1400, opened 2024-05-14
'Solution'

dsokolovskiy:


app/main.py
```diff
@@ -1,13 +1,27 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": [format_linter_error(error) for error in errors],
```
DevSpaciX: Look , you have the same logic implemented in format_single_linter_file function . Just reuse it insted of copying the same code ) And try always to stick DRY ( Do not repeat yourself ) principle : )

DevSpaciX (changes requested):

DevSpaciX (approved):
Good !

PR #1401, opened 2024-05-14
Solution_V1

BileichukIvan:


app/main.py
```diff
@@ -1,13 +1,45 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
```
DevSpaciX, line 7: Look , you have 2 places where this code can be reused . Always try to stick DRY ( Do not repeat yourself ) principle
Just reused ligic of format_linter_error function insted of copying same code : )

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            {
+                "line": error["line_number"],
```
DevSpaciX: Reuse existing logic here

app/main.py
```diff
@@ -1,13 +1,45 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": [
+                {
+                    "line": error["line_number"],
```
DevSpaciX: And reuse existing logic here too

DevSpaciX (changes requested):

BileichukIvan:
Like this?

DevSpaciX:
> Like this?

Almost

app/main.py
```diff
@@ -1,13 +1,30 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": [
```
DevSpaciX: Maybe better call format_single_linter_file here ? The same logic isn`t it ? DRY ))

DevSpaciX (changes requested):
Almost !

BileichukIvan:
Mb like this????

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(path, linter_report[path])
+        for path in linter_report.keys()
+    ]
```
Oleg-s-s: There is more convenient way to iterate over dictionary, especially when you need to use both keys and values

Oleg-s-s (changes requested):

Oleg-s-s (approved):
:+1:

PR #1402, opened 2024-05-14
Solution 1

SuskyiVolodymyr:


app/main.py
```diff
@@ -1,13 +1,34 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return (
+        {
+            "errors": [],
```
DevSpaciX: If the format_linter_error function does not find any errors, it will return an empty list as well. Therefore, you only need to adjust the part that may differ, specifically the 'status' logic.
Exp : 'status' : 'failed" if ... else ...

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1,13 +1,30 @@
...
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return (
+        {
+            "errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "failed"
+            if [format_linter_error(error) for error in errors]
+            else "passed"
```
Oleg-s-s: Why so complicated, you dont need to format error data to now that it exist

app/main.py
```diff
@@ -1,13 +1,30 @@
...
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(
+            error,
+            linter_report[error]
+        )
+        for error in linter_report
+    ]
```
Oleg-s-s: there is a better way to work with dictionaries in loops

Oleg-s-s (changes requested):

app/main.py
Oleg-s-s: You ignored this comment?
you can simply check if there are some errors, no need to check through format_linter_error

Oleg-s-s (changes requested):

app/main.py
SuskyiVolodymyr: sorry, missed this comment

Oleg-s-s (approved):
:+1:

PR #1404, opened 2024-05-15
Solution

SrMusienko:


app/main.py
```diff
@@ -1,13 +1,33 @@
...
-    # write your code here
-    pass
+
+    return {
+        "name": [val for key, val in error.items() if key == "code"][0],
+        "line": [val for key, val in error.items() if key == "line_number"][0],
+        "column":
+            [val for key, val in error.items() if key == "column_number"][0],
+        "message": [val for key, val in error.items() if key == "text"][0],
+        "source": "flake8"
```
Oleg-s-s: read about dict.get() or at least simplify this, looks too complicated

app/main.py
```diff
@@ -1,13 +1,33 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [format_linter_error(i) for i in errors],
```
Oleg-s-s: i is not good.
We use 1 symbol names only if its a value used to track iteration
We use _ if value if not used
We name normally in every other cases

app/main.py
```diff
@@ -1,13 +1,33 @@
...
-    pass
+
+    return [
+        format_single_linter_file(key, linter_report[key])
+        if bool(linter_report.get(key))
+        else {
+            "errors": [],
+            "path": key,
+            "status": "passed"
+        }
```
Oleg-s-s: No need in this part, format_single_linter_file should handle this

app/main.py
```diff
@@ -1,13 +1,33 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [format_linter_error(i) for i in errors],
+        "path": file_path,
+        "status": "failed"
```
Oleg-s-s: This function should handle both cases

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1,13 +1,26 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [format_linter_error(err_point) for err_point in errors],
+        "path": file_path,
+        "status": "failed" if bool(errors) else "passed"
```
Oleg-s-s: almost it, there is no need to specify `bool`. When you use expression `if value:` it will automatically return bool value. True if value exist and not null and False if opposite

app/main.py
```diff
@@ -1,13 +1,26 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [
+        format_single_linter_file(key, linter_report[key])
+        for key in linter_report.keys()
```
Oleg-s-s: I case when you are operating with both key and value of dict items it better to iterate over key and value. It will be more clear that you use both values

Oleg-s-s (changes requested):

Oleg-s-s (approved):

PR #1407, opened 2024-05-15
Solution

Unlie9:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if not errors else "failed"
```
Oleg-s-s, line 16: Why complicate it? 
```suggestion
        "status": "failed" if errors else "passed"
```

Oleg-s-s (approved):
:+1:

app/main.py
Unlie9: I will take this point into account and in the future I will not complicate things

> Why complicate it?

PR #1409, opened 2024-05-15
Solution

Irina17191:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "error": [format_linter_error(error=error) for error in errors],
```
Oleg-s-s: here is why test are not passed))
key should be errors, not error

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
Oleg-s-s: rename key, value to more descriptive ones

Oleg-s-s (changes requested):

app/main.py
Irina17191: I got, Thank you so much!)

DevSpaciX (approved):
Good job!

PR #1422, opened 2024-05-20
task solution

Muskam09:


YuriiSmolii (changes requested):
Reuse function `format_single_linter_file` inside `format_linter_report` function.

PR #1444, opened 2024-06-02
solution

ArtemShakun:


app/main.py
```diff
@@ -1,13 +1,24 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path, errors)
+        if errors else {"errors": [], "path": file_path, "status": "passed"}
+        for file_path, errors in linter_report.items()
```
viktoria-rybenchuk: DRY. You should reuse previous created function

viktoria-rybenchuk (changes requested):

app/main.py
ArtemShakun: > DRY. You should reuse previous created function

@viktoria-rybenchuk, Hey.
Sorry, but I don't understand this comment.  In line 21 I reused the function format_single_linter_file and in the loop, I got "file_path", and "errors" for this function. Is that right?

app/main.py
```diff
@@ -1,13 +1,24 @@
...
+        "path": file_path, "status": "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path, errors)
+        if errors else {"errors": [], "path": file_path, "status": "passed"}
```
viktoria-rybenchuk: It's better to define  this logic in ` format_single_linter_file`  and then  you can avoid checking `if errors`  and defining dict 2 times.

ArtemShakun: Thanks for the explanation.
Fixed

EdAlekseiev (approved):

PR #1453, opened 2024-06-09
Develop

HalynaRohatska:


app/main.py
```diff
@@ -1,13 +1,24 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(item) for item in errors],
+        "path": file_path,
+        "status": ("passed" if len(errors) == 0 else "failed"),
```
MNDonut: One small note: In python you can use just "if errors" or "if not errors" to check the list length. "if errors" returns True if there are some elements inside this list, otherwise False
Also brackets are redundant here as there is no "and" "or" operators

MNDonut (changes requested):

app/main.py
HalynaRohatska: added needed changes

app/main.py
```diff
@@ -12,7 +12,7 @@ def format_single_linter_file(file_path: str, errors: list) -> dict:
     return {
         "errors": [format_linter_error(item) for item in errors],
         "path": file_path,
-        "status": ("passed" if len(errors) == 0 else "failed"),
+        "status": "passed" if not errors else "failed",
```
MNDonut, line 15: it's easier to read "failed" if errors else "passed". Try to simplify if statements if possible :)

MNDonut (approved):

PR #1465, opened 2024-06-16
implemented linter formatter task for list and dict comprehensions

MasakDirt:


app/main.py
```diff
@@ -1,13 +1,41 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "column": error["column_number"], "line": error["line_number"],
+        "message": error["text"], "name": error["code"], "source": "flake8"
```
MNDonut: The best practice is to move each key on a separate line

app/main.py
```diff
@@ -1,13 +1,41 @@
...
-    # write your code here
-    pass
+    return {
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
```
MNDonut: Note that basically it's what format_linter_error makes. It formats in this way, so you can re-use format_linter_error function

app/main.py
```diff
@@ -1,13 +1,41 @@
...
+            } for error in errors
+        ],
+        "path": file_path, "status": "failed"
```
MNDonut: What about "passed" status? It's possible that there are no errors for same file, thus status should be "passed". It's mentioned in the requirements ;)

app/main.py
```diff
@@ -1,13 +1,41 @@
...
+                    "name": error["code"],
+                    "source": "flake8"
+                }
+                for error in linter_report[reports]
+            ],
+            "path": reports,
+            "status": "passed" if len(linter_report[reports]) == 0
+            else "failed"
+        }
+        for reports in linter_report
```
MNDonut: A few notes:
1. When you loop through dict, you get keys without their values. You can use dict.items() and it returns (key, value). In this way you can loop like:
for key, value in dict.items(): ...
2. Using #1 you can simplify linter_report[reports] statements
3. Using #1 you have values and can simplify if statement that you use for "status" key
4. Also note that this formatting is already implemented in your functions, so you can re-use it

MNDonut (changes requested):
Some improvements can be made ;)

app/main.py
MasakDirt: I understand and have refactor my code)))

I agree with all improvements, don't know why I don't seen it before myself, yet.)

MNDonut (approved):
Now it looks better ;)

MasakDirt:
> Now it looks better ;)

I agree)

PR #1466, opened 2024-06-16
Solution

x0wer:


app/main.py
```diff
@@ -1,13 +1,119 @@
...
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
+
+
+def format_single_linter_file(file_path: str, errors: List[Dict[str, Any]]) \
+        -> Dict[str, Any]:
```
Dimosphen1: Do not use `\` for line breaking here and in the `format_linter_report` function as well. Place each argument on a new line instead

app/main.py
```diff
@@ -1,13 +1,119 @@
-def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+from typing import List, Dict, Any
 
 
-def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+errors = [
```
Dimosphen1: Remove this variable and value from this file, there is no need in it

Dimosphen1 (changes requested):

PR #1472, opened 2024-06-19
Develop

PavelTym:


app/main.py
```diff
@@ -1,13 +1,41 @@
...
+
+        format_single_linter_file(
+            list(linter_report.keys())[2],
+            linter_report[list(linter_report.keys())[2]]
+        ),
+
+        format_single_linter_file(
+            list(linter_report.keys())[3],
+            linter_report[list(linter_report.keys())[3]]
+        )
```
EdAlekseiev: You should  use cycle for formatting `linter_report `

EdAlekseiev (changes requested):

app/main.py
PavelTym: the task is fixed

viktoria-rybenchuk (approved):
Well done!

PR #1473, opened 2024-06-19
solution

poznianski:


app/main.py
```diff
@@ -1,13 +1,31 @@
...
-    pass
+    return {
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
+            } for error in errors
```
NazarSheremet: Use function `format_linter_error()`, exept code dublication

NazarSheremet (changes requested):
Need to optimize solution

NazarSheremet (approved):
Good job!

PR #1484, opened 2024-07-02
Implemented solution

dimak20:


app/main.py
```diff
@@ -1,13 +1,45 @@
...
-    pass
+    return {
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
+            }
```
MNDonut: note that format_linter_error function makes this format, so you can use your already implemented function ;)

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+            for error in errors
+        ],
+        "path": file_path,
+        "status": "failed" if len(errors) != 0 else "passed"
```
MNDonut: In python you can use just "if errors" "if not errors" and python checks: if there are some elements in this list, it'll return True, otherwise False. This approach also works

app/main.py
```diff
@@ -1,13 +1,45 @@
...
-    pass
+    return [
+        {
+            "errors": [
+                {
+                    "line": error["line_number"],
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8"
```
MNDonut: note that you can use already implemented function for this formatting

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+                }
+                for error in value
+            ],
+            "path": key,
+            "status": "failed" if len(value) != 0 else "passed"
```
MNDonut: note from the function #2

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+        }
+        for key, value in linter_report.items()
```
MNDonut: I would recommend to use more explicit names, like:
for file_path, errors in linter_report.items(). It makes code more readable for other developers ;)

MNDonut (changes requested):

MNDonut (approved):
Perfect solution!

PR #1496, opened 2024-07-08
Solution_v1.0

Rozdorozhnii:


app/main.py
```diff
@@ -1,13 +1,26 @@
...
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error)
+            for error in errors
+            if error["filename"] == file_path],
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed"
```
LiudmylaKulinchenko: You can just check `if errors` here

app/main.py
```diff
@@ -1,13 +1,26 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error)
+            for error in errors
+            if error["filename"] == file_path],
```
LiudmylaKulinchenko: fix formatting here, please
Both brackets should be on the same level, u can see checklist

LiudmylaKulinchenko (changes requested):

NazarSheremet (approved):
Good job!

PR #1498, opened 2024-07-11
Solution

OleksiukStepan:
Implement methods:
- format_linter_error
- format_single_linter_file
- format_linter_report


MNDonut:
Actually it's functions, not methods ;)
Method is a function inside a class

MNDonut (approved):

PR #1500, opened 2024-07-11
Solution

ihorutkin:


app/main.py
```diff
@@ -1,13 +1,47 @@
...
-    # write your code here
-    pass
+    return {
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8",
```
MNDonut: note that this is basically what format_linter_error does, so you can reuse this function

app/main.py
```diff
@@ -1,13 +1,47 @@
...
+            }
+            for error in errors
+        ],
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed"
```
MNDonut: In python you can just do "if errors", "if not errors". Python will automatically check if there is at least 1 element in this list. If use -> it'll be True, otherwise False

app/main.py
```diff
@@ -1,13 +1,47 @@
...
-    pass
+    return [
+        {
+            "errors": [] if len(linter_report[f"{report}"]) == 0 else [
+                {
+                    "line": error["line_number"],
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8",
```
MNDonut: code duplicate, you can reuse already implemented function :)

app/main.py
```diff
@@ -1,13 +1,47 @@
...
+        "status": "passed" if len(errors) == 0 else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": [] if len(linter_report[f"{report}"]) == 0 else [
```
MNDonut: you can simplify this if statement as I mentioned above

app/main.py
```diff
@@ -1,13 +1,47 @@
...
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8",
+                }
+                for error in linter_report[f"{report}"]
+            ],
+            "path": report,
+            "status": "passed" if
+            len(linter_report[f"{report}"]) == 0
+            else "failed"
```
MNDonut: you can use .items() method of dict that returns (key, value) tuples, thus you can iterate through keys and values at the same time instead of fetching value for each key every time

MNDonut (changes requested):

oliasymak24 (approved):
You changed everything correct, good job!

PR #1501, opened 2024-07-12
Write 3 functions

A-n-a-s-t-a-s-i-i-a:


app/main.py
```diff
@@ -1,13 +1,17 @@
...
+            "name": error["code"],
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "failed" if len(errors) > 0 else "passed"}
```
MNDonut: By PEP8 keys and values should start from the next line

MNDonut: In Python you can just use "if errors". Python automatically checks if there are some values in errors list. If there are - it'll return True, otherwise False

app/main.py
```diff
@@ -1,13 +1,17 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(file_path, errors)
+            for file_path, errors in linter_report.items()]
```
MNDonut: When you use brackets and line length > 79 characters, it's better to do it in this way
```
return [
    format_single_linter_file(file_path, errors)
    for file_path, errors in linter_report.items()
]
```

MNDonut (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if len(errors) else "passed"
```
oliasymak24: You don't need additionally call len() function.

oliasymak24 (approved):
You changed all the comments correctly but pay attention to the small comment

PR #1502, opened 2024-07-13
all three methods format correctly

Narberal90:


app/main.py
```diff
@@ -1,13 +1,25 @@
...
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(formatted_dict) for formatted_dict in errors
+        ],
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed"
```
MNDonut, line 17: you can use just "if errors" or "if not errors". Python automatically checks if there are some objects in this list, it'll return True, otherwise False

MNDonut (approved):

PR #1505, opened 2024-07-14
solution

BastovOleksandr:


app/main.py
```diff
@@ -1,13 +1,28 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error)
+            for error in errors
```
MNDonut: If line length < 79 characters there is no need to split it into 2 lines

app/main.py
```diff
@@ -1,13 +1,28 @@
...
+        ],
+        "path": file_path,
+        "status":
+            "failed" if len(errors)
+            else "passed"
```
MNDonut: If line length < 79 characters there is no need to split it into 2 lines

MNDonut (changes requested):

oliasymak24 (approved):
Good job!

PR #1506, opened 2024-07-15
implement py-linter-formatter

ValeriiaKyr:


app/main.py
```diff
@@ -1,13 +1,44 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"}
```
MNDonut: move that bracket on the next line

app/main.py
```diff
@@ -1,13 +1,44 @@
...
-    # write your code here
-    pass
+    return {
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
```
MNDonut: It seems like you have already implemented function for it ;)

app/main.py
```diff
@@ -1,13 +1,44 @@
...
+            }
+            for error in errors
+        ],
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
```
MNDonut: in python you can just use "if errors". Python will check if there are some objects in errors

app/main.py
```diff
@@ -1,13 +1,44 @@
...
-    pass
+    return [
+        {
+            "errors": [
+                {
+                    "line": error["line_number"],
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8"
```
MNDonut: seems like you have already implemented function for this ;)

app/main.py
```diff
@@ -1,13 +1,44 @@
...
+                }
+                for error in linter_report[path]
+            ],
+            "path": path,
+            "status": "failed" if linter_report[path] else "passed"
+        }
+        for path in linter_report
```
MNDonut: it order to not to do `linter_report[path]` every time, you can use .items() method of dict so that you'll be able to loop through keys and values at the same time ;)

MNDonut (changes requested):

MNDonut (approved):

PR #1510, opened 2024-07-16
Solution

vladislav-tsybuliak1:


app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(file, errors)
+            for file, errors
+            in linter_report.items()]
```
oliasymak24: in this case you are not exceeding the 79 character limit, so it is better not to split the loop (`for file, errors
            in linter_report.items()` )

oliasymak24 (changes requested):
Your code looks good, some small comment.

oliasymak24 (approved):
Good job!

PR #1511, opened 2024-07-17
My excellent solution of py-linter-formatter

AndriiHamasa:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(obj) for obj in errors],
+        "path": file_path,
+        "status": "failed" if len(errors) != 0 else "passed"
```
oliasymak24: You don't need to check `len(errors)`, just `if errors` will return the same result.

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(obj, linter_report[obj])
+        for obj in linter_report
```
oliasymak24: Try to use for iteration linter_report.items()

oliasymak24 (changes requested):
In general your code works good, but look into some small comments

MNDonut (approved):

PR #1515, opened 2024-07-17
Solution

mnezbrytska:


app/main.py
```diff
@@ -1,13 +1,29 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [{"line": error["line_number"],
+                        "column": error["column_number"],
+                        "message": error["text"],
+                        "name": error["code"],
+                        "source": "flake8"} for error in errors],
```
LiudmylaKulinchenko: You are repeating yourself here. Just use the function that you create above. Don't duplicate your code.

app/main.py
```diff
@@ -1,13 +1,29 @@
...
+    return [
+        {"errors": [{"line": error["line_number"],
+                     "column": error["column_number"],
+                     "message": error["text"],
+                     "name": error["code"],
+                     "source": "flake8"} for error in errors],
+         "path": file_path,
+         "status": "passed" if not errors else "failed"}
+        for file_path, errors in linter_report.items()
+    ]
```
LiudmylaKulinchenko: same here

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1,13 +1,19 @@
...
+            "status": "passed" if not errors else "failed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path, errors)
+        for file_path, errors in linter_report.items()
+    ]
```
MNDonut: Note how you place brackets for `format_linter_report` function(correct way) and for other functions :)

app/main.py
```diff
@@ -1,13 +1,19 @@
...
+            "name": error["code"],
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "passed" if not errors else "failed"}
```
MNDonut: It's easier to read if you do `... if errors ...`. No need to do vice-versa if statement when you can do a direct condition

MNDonut (changes requested):

app/main.py
```diff
@@ -1,13 +1,17 @@
...
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "failed" if errors else "passed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(file_path, errors)
+            for file_path, errors in linter_report.items()]
```
MNDonut: let's agree to use the first good example from checklist (1 tabulation and code should be from the next line)
image

MNDonut (changes requested):

LiudmylaKulinchenko (approved):
Well done 🔥

PR #1517, opened 2024-07-17
'Solution'

Ivan-Shakhman:
Results of task

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -12,10 +12,10 @@ def format_single_linter_file(file_path: str, errors: list) -> dict:
     return {
         "errors": [format_linter_error(error) for error in errors],
         "path": file_path,
-        "status": "passed" if errors == [] else "failed",
+        "status": "passed" if not errors else "failed",
     }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    return [format_single_linter_file(key, value)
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if errors == [] else "failed",
```
LiudmylaKulinchenko: You can just check `if errors` here. It will cover cases when errors is None or an empty list

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(key, value)
+            for key, value in linter_report.items()]
```
LiudmylaKulinchenko: Use descriptive names, instead of key/value here

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks great! 🌸

PR #1519, opened 2024-07-17
Solution

panichqa:


app/main.py
```diff
@@ -1,13 +1,22 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if [format_linter_error(error)
+                               for error in errors] else "passed"
```
MNDonut: why do you need to make a new list comprehension where you call format_linter_error to check if there are errors?
This function just formats errors in an appropriate way, it doesn't decide whether there are errors or not. The source of errors is "errors" list that you get as a parameter

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(file_path, errors)
+            for file_path, errors in linter_report.items()]
```
MNDonut: it's better move this loop on the next line as you make with dicts in the other functions

MNDonut (changes requested):

app/main.py
panichqa: i did not understand what do you mean

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if not errors else "failed"
```
MNDonut: let's use `if errors` and swap strings, it'll be easier to read the code ;)

MNDonut (changes requested):

LiudmylaKulinchenko (approved):
Looks good 🔥

PR #1520, opened 2024-07-17
need help with 3rd task "format_linter_report"

Alexandr-Politov:
need help with 3rd task "format_linter_report"

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,21 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
-
+    return {
+    "line": error["line_number"],
+    "column": error["column_number"],
+    "message": error["text"],
+    "name": error["code"],
+    "source": "flake8"
+    }
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
-
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed"
+    }
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path, errors)
+        for file_path, errors in format_single_linter_file(file_path, errors)
+    ]
--- a/app/main.py
+++ b/app/main.py
@@ -1,21 +1,21 @@
 def format_linter_error(error: dict) -> dict:
     return {
-    "line": error["line_number"],
-    "column": error["column_number"],
-    "message": error["text"],
-    "name": error["code"],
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,21 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
-
+    return {
+    "line": error["line_number"],
+    "column": error["column_number"],
+    "message": error["text"],
+    "name": error["code"],
+    "source": "flake8"
```
MNDonut: don't forget about identation

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
-
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed"
```
MNDonut: but here you have a correct identation :)

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file_path, errors)
+        for file_path, errors in format_single_linter_file(file_path, errors)
```
MNDonut: Something is definitely wrong because of 3 reasons:
1. you don't use your input parameter `linter_report`
2. `format_single_linter_file` returns 1 dict and you cannot iterate through just dict using 2 variables **file_path** and **errors**
3. At the end you call `format_single_linter_file(file_path, errors)`. But where do you take file_path and errors? Your function doesn't have these parameters, you have only **linter_report**

So these issues are interdependent. When you fix one, the others are resolved :)

MNDonut (changes requested):

MNDonut (changes requested):
Tests fail because of code style issues

image

app/main.py
```diff
@@ -1,13 +1,26 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
```
LiudmylaKulinchenko: You can just check `if errors` here, it will cover None and an empty list

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job 🔥

PR #1521, opened 2024-07-17
exercise-errors

yashkunn:


app/main.py
```diff
@@ -1,13 +1,25 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
     pass
```
MNDonut: You can remove this redundant pass ;)

app/main.py
```diff
@@ -1,13 +1,25 @@
...
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if [format_linter_error(error)
+                               for error in errors] else "passed"
+    }
     pass
```
MNDonut: You can remove this redundant pass ;)

app/main.py
```diff
@@ -1,13 +1,25 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
+    return [format_single_linter_file(file_path, errors)
+            for file_path, errors in linter_report.items()]
     pass
```
MNDonut: You can remove this redundant pass ;)

app/main.py
```diff
@@ -1,13 +1,25 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if [format_linter_error(error)
+                               for error in errors] else "passed"
```
MNDonut: why do you need to make a new list comprehension where you call `format_linter_error` to check if there are errors?
This function just formats errors in an appropriate way, it doesn't decide whether there are errors or not. The source of errors is "errors" list that you get as a parameter

app/main.py
```diff
@@ -1,13 +1,25 @@
...
+    }
     pass
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
+    return [format_single_linter_file(file_path, errors)
+            for file_path, errors in linter_report.items()]
```
MNDonut: it's better move this loop on the next line as you make with dicts in the other functions

MNDonut (changes requested):

app/main.py
```diff
@@ -1,13 +1,26 @@
...
 
+# checks for list content and sets the appropriate status
+# and handles errors with the first method
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if errors else "passed"}
```
MNDonut: move the bracket on the next line

MNDonut (changes requested):

LiudmylaKulinchenko (approved):
Well done 🌸

PR #1522, opened 2024-07-18
Solution

ABaidin:


app/main.py
```diff
@@ -1,13 +1,33 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [{"line": error["line_number"],
+                        "column": error["column_number"],
+                        "message": error["text"],
+                        "name": error["code"],
+                        "source": "flake8"}
```
oliasymak24: You have double code, try just call function which you created before.

app/main.py
```diff
@@ -1,13 +1,33 @@
...
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {"errors": [{"line": error["line_number"],
+                     "column": error["column_number"],
+                     "message": error["text"],
+                     "name": error["code"],
+                     "source": "flake8"}
```
oliasymak24: You have double code, try just call function which you created before.

app/main.py
```diff
@@ -1,13 +1,33 @@
...
+                    for error in value],
+         "path": key,
+         "status": "failed" if value
+         else "passed"}
```
oliasymak24: The same problem here, you already have this check in other function.

oliasymak24 (changes requested):
In general looks correct, but a lot of duplicates.

ABaidin:
> In general looks correct, but a lot of duplicates.

Thank you, I didn't even think of calling previous functions. I treated these 3 functions like different tasks. 
Thanks again for your advice, it should look better now :)

app/main.py
```diff
@@ -1,13 +1,20 @@
...
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "failed" if errors
+            else "passed"}
```
MNDonut: let's follow the first good example from checklist(like you did for the list in format_linter_report)

MNDonut (changes requested):

app/main.py
```diff
@@ -1,13 +1,24 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if errors
+        else "passed"
```
LiudmylaKulinchenko: You shouldn't move `else` to the next line 👀

app/main.py
```diff
@@ -1,13 +1,24 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
LiudmylaKulinchenko: Also you can use more descriptive names here instead of key, value!

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Great job! 🥳

PR #1550, opened 2024-07-30
Solution for py-linter-formatter

nikvitkaua:


app/main.py
```diff
@@ -1,13 +1,46 @@
...
+        "errors":
+            [
+                {
+                    "line": error.get("line_number"),
+                    "column": error.get("column_number"),
+                    "message": error.get("text"),
+                    "name": error.get("code"),
+                    "source": "flake8"
+                }
+                for error in errors
```
EdAlekseiev: You should use `format_linter_error` here to avoid code duplication

app/main.py
```diff
@@ -1,13 +1,46 @@
...
+            ],
+            "path": path,
+            "status": "failed" if errors else "passed"
```
EdAlekseiev: You should use `format_single_linter_file` to avoid code duplication

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1,13 +1,34 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": [
+                format_linter_error(error)
```
vasylhnatiuk: reuse format_single_linter_file here

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1551, opened 2024-07-31
'Solution'

PonomaryovNikita:


app/main.py
```diff
@@ -1,13 +1,36 @@
...
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [{"line": error["line_number"],
```
pavlopro: You should utilize `format_linter_error` method here

app/main.py
```diff
@@ -1,13 +1,36 @@
...
+
+            }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": [format_linter_error(error) for error in errors],
```
pavlopro: You should utilize `format_single_linter_file` method here

pavlopro (changes requested):
Please review realization of the task

app/main.py
```diff
@@ -1,13 +1,20 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {"line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"}
```
Dimosphen1: Place all keys on a new line so that there would be no keys on the same line with `{` and `}` here and in the `format_single_linter_file` function as well to improve the readability

Dimosphen1 (changes requested):
Several minor changes were requested

vsmutok (approved):
Good Job!

PR #1553, opened 2024-08-03
Solution

EvhYa:


app/main.py
```diff
@@ -1,13 +1,25 @@
...
-    pass
+    return {
+        key: value for key, value in {
+            "line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"
+        }.items()
+    }
```
vsmutok:
```suggestion
    return {
        "line": error["line_number"],
        "column": error["column_number"],
        "message": error["text"],
        "name": error["code"],
        "source": "flake8"
    }
```
no need to use loop here

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1554, opened 2024-08-04
My solution

lavryniuk-dev:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed",
```
LiudmylaKulinchenko: `if errors` will cover cases if the list is blank and if errors=None
No need to get length in this case

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🌞

PR #1561, opened 2024-08-07
Solution

arsen-arutiunov:


app/main.py
```diff
@@ -1,13 +1,127 @@
+errors = [
```
vasylhnatiuk: why do you need this??

vasylhnatiuk (changes requested):
fix tests

app/main.py
arsen-arutiunov: Forgot to delete it, I'll fix it now.

PR #1584, opened 2024-08-17
Solution

KostiantinPastukh:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if len(errors) != 0 else "passed"
```
```suggestion
        "status": "failed" if errors else "passed"
```

viktoria-rybenchuk: no need to check with 0

viktoria-rybenchuk (changes requested):

app/main.py
KostiantinPastukh: I fixed, thank you

vsmutok (approved):
Well Done!

PR #1592, opened 2024-08-23
Format Linter Errors

Hedgehog18:
Format Linter Errors


EdAlekseiev (changes requested):
You should fix flake8 errors

Hedgehog18:
Final update

PR #1596, opened 2024-08-27
"Solution"

Olha-y:


app/main.py
```diff
@@ -1,13 +1,26 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
+    return {
+        "line": error.get("line_number"),
+        "column": error.get("column_number"),
+        "message": error.get("text"),
+        "name": error.get("code"),
+        "source": "flake8",
+    }
     pass
```
vasylhnatiuk: delete pass

vasylhnatiuk: do it everywhere

vasylhnatiuk (changes requested):

app/main.py
Olha-y: deleted pass everywhere

vasylhnatiuk (approved):

PR #1616, opened 2024-09-08
Solution-py-linter-formatter

skyfoxwork:


app/main.py
```diff
@@ -1,13 +1,24 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
+    ]
```
Nattalli: don't use `key, value` if you already know field names. for example `file_path, errors` would be better

app/main.py
```diff
@@ -1,13 +1,24 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        ("name" if key == "code" else "source" if key == "filename" else
+         "message" if key == "text" else tuple(key.split("_"))[0]):
+        ("flake8" if key == "filename" else value)
+        for key, value in error.items() if key != "physical_line"
```
Nattalli: not readable code. please, use static key names and choose the correct fields from the dictionary for those keys. for example:
```python
"name": error["code"]
```

app/main.py
```diff
@@ -1,13 +1,24 @@
...
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        ("errors" if elem == "error" else elem):
+        ([format_linter_error(error) for error in errors]
+         if elem == "error" else file_path if elem == "path" else
+         "failed" if elem == "status" and errors != [] else "passed")
+        for elem in ("error", "path", "status")
```
Nattalli: the same here. also, `elem` is bad naming for variable

Nattalli (changes requested):
comments added

Nattalli (approved):
👍🏻

PR #1624, opened 2024-09-11
'Solution'

spa7id:


app/main.py
```diff
@@ -1,13 +1,22 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {new_key: error[old_key] for new_key, old_key in {
+        "line": "line_number",
+        "column": "column_number",
+        "message": "text",
+        "name": "code"
+    }.items()} | {"source": "flake8"}
```
Nattalli: it seems difficult, and not readable. Do you need to iterate over fields? You can do the same as in the next function

Nattalli (changes requested):
improve first function, make it readable and easy to understand

Nattalli (approved):
👍🏻

PR #1637, opened 2024-09-16
implement py-linter-formatter

94nj111:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "name": error.get("code"),
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": ([format_linter_error(error) for error in errors]),
```
Nattalli: not a mistake, but you don't need to use additional braces `()`- list comprehension is good enough without them

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "path": file_path,
+        "status": ("failed" if errors else "passed")
```
Nattalli: the same with braces here

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(path, linter_report.get(path))
+            for path in linter_report]
```
Nattalli: this code is correct but a little bit complex to understand. I believe that using `.items()` can make it more readable, what do you think?

Nattalli: also, I don't like that you use 2 different code styles in 1 file. the example from guideline:

Знімок екрана 2024-09-18 о 15 02 25

both examples are good but use one:)

Nattalli (changes requested):
comments added

app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(*path) for path in linter_report.items()
```
Nattalli: is it correct to name two items just a `path`? Probably it. would be better to use 2 variables and give them correct name, where first would be really a path

Nattalli (changes requested):

94nj111:
maybe this name would be better?

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(*path_and_errors)
+        for path_and_errors in linter_report.items()
```
Y-Havryliv, line 22: it would be better to use two variables, smth like this:
`for file, report in linter_report.items()`
but ur solution is also ok

Y-Havryliv (approved):
gj!

PR #1638, opened 2024-09-16
Solution

zaietsmo:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file, errors)
+        for file, errors in linter_report.items()
```
Nattalli: `file_path` would be better

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(err) for err in errors],
```
Nattalli: `error` word and `err` do have not a big difference. Does it need to be shortened?

Nattalli (changes requested):
also I recommend you use one of the following styles in one file: 
Знімок екрана 2024-09-18 о 15 40 02

AnyoneClown (approved):
Good job!

PR #1641, opened 2024-09-17
Solution

sashasyrota:


app/main.py
```diff
@@ -1,13 +1,28 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
+    return {"line": error["line_number"], "column": error["column_number"],
+            "message": error["text"], "name": error["code"],
+            "source": "flake8"} if len(error) > 0 else {}
```
LLkaia:
```suggestion
    return {
        "line": error["line_number"],
        "column": error["column_number"],
        "message": error["text"],
        "name": error["code"],
        "source": "flake8"
    }
```

Please, place one item per line for this and following cases

app/main.py
```diff
@@ -1,13 +1,28 @@
...
     pass
```
LLkaia: Remove this `pass` for this and following appearances

app/main.py
```diff
@@ -1,13 +1,28 @@
...
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
+    return {"errors":
+            [{"line": error["line_number"],
+              "column": error["column_number"],
+              "message": error["text"], "name": error["code"],
+              "source": "flake8"}
+             if len(error) > 0 else {} for error in errors],
+            "path": file_path, "status": "failed"
+            if len(errors) > 0 else "passed"}
```
LLkaia: Use function above instead of duplicating same code

app/main.py
LLkaia: No reason for length check in this task

app/main.py
```diff
@@ -1,13 +1,28 @@
...
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
+    return [{"errors":
+            [{"line": error["line_number"], "column": error["column_number"],
+              "message": error["text"], "name": error["code"],
+              "source": "flake8"}
+             if len(error) > 0 else {} for error in errors],
+            "path": name, "status": "failed" if len(errors) > 0 else "passed"}
+            for name, errors in linter_report.items()]
```
LLkaia: same here

LLkaia (changes requested):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed"
+        if len(errors) > 0 else "passed"
```
LLkaia:
```suggestion
        "status": "failed" if errors else "passed"
```

app/main.py
```diff
@@ -1,13 +1,23 @@
...
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    } \
+        if len(error) > 0 else {}
```
LLkaia:
```suggestion
    }
```

Since we are not expecting that someone will call this function with wrong data. Otherwise, you need to check if all keys exist in this dict, but this is not needed to be implemented in scope of this task.

LLkaia (changes requested):

app/main.py
```diff
@@ -1,13 +1,26 @@
...
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+        if
+        len(error) > 0 and "line_number" in error.keys()
+        and "column_number" in error.keys()
+        and "text" in error.keys() and "code" in error.keys()
+        else {}
```
viktoria-rybenchuk: no need make this checking

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(file_path=key, errors=errors)
+            for key, errors in linter_report.items()]
```
pavlopro: Please use proper variable names. Instead of `key` use something like `report_path`

pavlopro (changes requested):

vsmutok (approved):
Good Job!

PR #1663, opened 2024-09-23
Solution

Den-k0:


dxrrkwm (approved):
well done😇

app/main.py
```diff
@@ -1,13 +1,48 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
```
Y-Havryliv, line 13: it's better to reuse `format_linter_error` to format errors

app/main.py
```diff
@@ -1,13 +1,48 @@
...
+        ],
+        "path": file_path,
+        "status": "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
```
Y-Havryliv, line 23: and here you can reuse `format_single_linter_file`

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1,13 +1,26 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
```
AnyoneClown: Use `get()` method for dict, in case if you don't have some `key` inside of your dict and you will try to get it with `[key]` - it will give you error

app/main.py
```diff
@@ -1,13 +1,26 @@
...
+        "path": file_path,
+        "status": "passed" if not errors else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(filepath, issues)
```
AnyoneClown: Python is using snake_case. So, it should be `file_path`

AnyoneClown (changes requested):

Nattalli (approved):
👍🏻

PR #1666, opened 2024-09-24
Develop

Rabb1t-Artem:


sumseer (commented):
gj!

sumseer (approved):

app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "status": "failed" if errors else "passed",
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(file, errors)
+        for file, errors in linter_report.items()
```
AnyoneClown: Better to name it `file_path` to be more descriptive

app/main.py
```diff
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8",
+    }
```
AnyoneClown: Use `get()` method for dict, in case if you don't have some key inside of your dict and you will try to get it with `[key]` - it will give you error

AnyoneClown (changes requested):

Nattalli (approved):
👍🏻

PR #1668, opened 2024-09-24
Solution

sumseer:


MrEgorLite:
:+1:

MrEgorLite (approved):

app/main.py
```diff
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
```
AnyoneClown: Use `get()` method for dict, in case if you don't have some key inside of your dict and you will try to get it with `[key]` - it will give you error

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job!

PR #1670, opened 2024-09-24
Solution v1

MrEgorLite:


SerhiiK05:
Actually you don't need brackets around dict after return and next time don't forget to delete "pass", so it's better to read your code. Well Done and GJ!

alauda6 (approved):
Actually you don't need brackets around dict after return and next time don't forget to delete "pass", so it's better to read your code. Well Done and GJ!

alauda6 (approved):
Good job!

app/main.py
```diff
@@ -1,13 +1,20 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
+    return ({"line": error.get("line_number"),
+             "column": error.get("column_number"),
+             "message": error.get("text"),
+             "name": error.get("code"),
+             "source": "flake8"})
     pass
```
Nattalli: remove `pass` because all code after `return` statement wouldn't be executed

app/main.py
```diff
@@ -1,13 +1,20 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
+    return ({"errors": [format_linter_error(error) for error in errors],
+             "path": file_path,
+             "status": "passed" if errors == [] else "failed"})
     pass
```
Nattalli: the same here

app/main.py
```diff
@@ -1,13 +1,20 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+
+    return [format_single_linter_file(file_path, errors)
+            for file_path, errors in linter_report.items()]
```
Nattalli, line 18: and here

app/main.py
```diff
@@ -1,13 +1,20 @@
...
+             "name": error.get("code"),
+             "source": "flake8"})
     pass
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
+    return ({"errors": [format_linter_error(error) for error in errors],
+             "path": file_path,
+             "status": "passed" if errors == [] else "failed"})
```
Nattalli: checking `== []` is redundant
```suggestion
             "status": "passed" if errors else "failed"})
```

Nattalli (changes requested):

app/main.py
Nattalli: @MrEgorLite sorry, that should be:

```suggestion
             "status": "passed" if not errors else "failed"})
```

Nattalli (changes requested):

Nattalli (approved):
perfect 💅🏻

PR #1671, opened 2024-09-25
'Solution'

AndriiDementiev:


MatvijSafonov (approved):

app/main.py
```diff
@@ -1,13 +1,26 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(element, linter_report[element])
+        for element in linter_report
+    ]
```
Nattalli: not good. what is `element` and `linter_report[element]`? It is actually have be something like `file_path` and `errors`. You can use `.items()` to have 2 variables

app/main.py
```diff
@@ -1,13 +1,26 @@
...
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error)
+            for error in errors
+        ],
+        "path": file_path,
+        "status": "passed" if errors == [] else "failed"
```
Nattalli: just more professional and beautiful 🌹 💅🏻 

```suggestion
        "status": "passed" if not errors else "failed"
```

Nattalli (changes requested):
not bad but some comments added

Nattalli (approved):
👍🏻

PR #1674, opened 2024-09-25
Solution

levseenko:


VolodymyrSemchysyn (approved):
good job!
i think this code style will make it more readeble
                   return {
                               "line":  smth smth,
                               "column":  smth smth
                   }

Nattalli (approved):
👍🏻

Nattalli:
> good job! i think this code style will make it more readeble return { "line": smth smth, "column": smth smth }

probably yes but both solutions are good, as it mentioned in checklist

PR #1682, opened 2024-09-25
I can't passed 1 last test, I need help :(

Le0n-K:


app/main.py
```diff
@@ -1,13 +1,32 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "path": file_path,
+        "status": "failed" if errors else "success",
```
VladyslavBon: Here in else string sould be "passed", not "success"

app/main.py
```diff
@@ -1,13 +1,32 @@
+from __future__ import annotations
+
+from typing import Dict, List
+
+
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error.get("line_number"),
```
VladyslavBon, line 3: Here use the same syntax like in other values of this dict

app/main.py
```diff
@@ -1,13 +1,32 @@
+from __future__ import annotations
+
+from typing import Dict, List
```
VladyslavBon: You don't need any imports in this task

app/main.py
```diff
@@ -1,13 +1,32 @@
...
+        "errors": [format_linter_error(error) for error in errors]
+   }
 
 
-def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+def format_linter_report(linter_report: dict) -> dict:
+    return {
+        "path": linter_report.get("path"),
```
VladyslavBon: In this function you should only return previos function in nested list, don't need create and return another dict

VladyslavBon (approved):

app/main.py
Nattalli: yes, please, use only one style: `get()` or `[]` everywhere

app/main.py
```diff
@@ -1,13 +1,32 @@
...
+        "status": "success" if not linter_report.get("errors") else "failed",
+        "errors": [
+            format_single_linter_file(file["path"], file["errors"])
+            for file in linter_report.get("files", [])
+        ]
+    }
```
Nattalli: add an empty line in the end of the file

app/main.py
Nattalli: yes, use `format_single_linter_file` function instead

Nattalli (changes requested):

Y-Havryliv (approved):
nice!

PR #1687, opened 2024-09-29
Add linter formatter solution

charmheroku:


app/main.py
```diff
@@ -1,13 +1,25 @@
...
+        "status": "failed" if errors else "passed",
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, value)
+        for key, value in linter_report.items()
```
pavlopro: Use more appropriate variable names

pavlopro (changes requested):

app/main.py
charmheroku: Done

app/main.py
```diff
@@ -1,13 +1,25 @@
...
+        "status": "failed" if errors else "passed",
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(filename, list_errors)
+        for filename, list_errors in linter_report.items()
```
Dimosphen1, line 24: There is no need to indicate type in the variable name, avoid it in the future
```suggestion
        for filename, errors in linter_report.items()
```

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #1689, opened 2024-10-01
Solution

okdzeryn:


app/main.py
```diff
@@ -1,13 +1,27 @@
...
+    }
     pass
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
+    return [{"path": path,
+             "errors": [format_linter_error(error) for error in errors],
+             "status": "failed" if errors else "passed"
+             }
```
LLkaia: Use function `format_single_linter_file` instead

app/main.py
```diff
@@ -1,13 +1,27 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
     pass
```
LLkaia: Remove unnecessary `pass`

app/main.py
```diff
@@ -1,13 +1,27 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
     pass
```
LLkaia: same `pass`

app/main.py
```diff
@@ -1,13 +1,27 @@
...
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
+    return [{"path": path,
+             "errors": [format_linter_error(error) for error in errors],
+             "status": "failed" if errors else "passed"
+             }
+            for path, errors in linter_report.items()]
     pass
```
LLkaia: same `pass`

LLkaia (changes requested):

app/main.py
viktoria-rybenchuk: not fixed. Remove keyword - `pass`

app/main.py
viktoria-rybenchuk: the same here

app/main.py
viktoria-rybenchuk: and here

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #1694, opened 2024-10-07
Solution

serg121192:


app/main.py
```diff
@@ -1,13 +1,52 @@
...
+        "errors": [
+            {
+                "line": error["line_number"],
+                "column": error["column_number"],
+                "message": error["text"],
+                "name": error["code"],
+                "source": "flake8"
+            }
+            for error in errors
+        ],
```
LLkaia, line 25: Use function format_linter_error instead of duplicating your code

app/main.py
```diff
@@ -1,13 +1,52 @@
...
+                    "column": error["column_number"],
+                    "message": error["text"],
+                    "name": error["code"],
+                    "source": "flake8"
+                }
+                for error in linter_report[report]
+            ],
+            "path": report,
+            "status": "failed"
+        }
```
LLkaia, line 50: same here, but use  format_single_linter_file here

LLkaia (changes requested):

PR #1698, opened 2024-10-08
Updated solution

serg121192:


app/main.py
```diff
@@ -1,13 +1,30 @@
...
+        "path" : file_path,
+        "status" : "passed"
+    } if errors == [] else {
+        "errors" : [
+            format_linter_error(error)
+            for error in errors
+        ],
+        "path" : file_path,
+        "status" : "failed"
+    }
```
Dimosphen1: Please, use if condition in a value for `errors` key instead of forming two dicts here

Dimosphen1 (changes requested):

app/main.py
serg121192: But statuses at these dicts are different too. OR if double "if" condition like this:

    return {
        "errors" : [] if errors == [] else
        [
            format_linter_error(error)
            for error in errors
        ],
        "path" : file_path,
        "status" : "passed" if errors == [] else"failed"
    }
    
    is possible?

app/main.py
```diff
@@ -1,13 +1,27 @@
...
-    # write your code here
-    pass
+    return {
+        "errors" : [] if errors == [] else
+        [
+            format_linter_error(error)
+            for error in errors
+        ],
+        "path" : file_path,
+        "status" : "passed" if errors == [] else "failed"
```
viktoria-rybenchuk: you can check if errors, no need to compare with empty list

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1,13 +1,27 @@
...
+        "name" : error["code"],
+        "source" : "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors" : [] if not errors else
```
pavlopro: Pplease, simplify this to `[format_linter_error(error) for error in errors]`

pavlopro (changes requested):

pavlopro (approved):
GJ!

PR #1715, opened 2024-10-15
Solution

dottfmar:


4ikkamony (approved):
Everything is in order. Formatting is up to the standart, and all the tests have passed.

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+            "path": file_path,
+            "status": "failed" if errors else "passed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        format_single_linter_file(key, linter_report[key])
+        for key in linter_report.keys()
```
4ikkamony, line 20: Just one small thing:
specifying .keys() here is redundant, as Python iterates through keys by default.
This is not an issue, of course.
In my opinion, specifying things explicitly is the way to go, but it's good to know when you can save yourself a keystroke or two ;)

PR #1746, opened 2024-10-26
Solution

bodiakof:


L1nk3rrr (changes requested):
Well done!
From the code and logical side, all is good, but from formatting, I found 2 minor points to fix in this theme.

We don't need to newline after the name of the function(all your 3 functions have it)
Don't forget about newline at the end of a file.

L1nk3rrr (approved):
You forget about comment with a blank line at the end of a file, for now I approve it, but next time will be returned to you, anyway you did task and improve main style point that commented previously

PR #1750, opened 2024-10-28
Solution

Chukhno-Maksym:


app/main.py
```diff
@@ -1,13 +1,26 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed"}
```
petrykivd, line 15: You need to follow the same style code everywhere, so if you want brackets on separate lines, it should be like this at the beginning and at the end 😄

petrykivd (approved):
Good job! 🚀

PR #1787, opened 2024-11-13
Solution to py-linter-formatter task

Trynadis:
Written 3 functions to convert the report into the right format:
- format_linter_error - formats a single error
- format_single_linter_file - formats all errors for a particular file and adds the path key — path to the file, and the status key — "failed" if there are errors, "passed" if there are no errors
- format_linter_report - formats all errors for all report files


Alkaponees (approved):
LGMT!

PR #1795, opened 2024-11-15
Solution

maxakim0v:


app/main.py
```diff
@@ -1,13 +1,22 @@
...
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(file_path, errors)
+            for file_path, errors in linter_report.items()
+            ]
```
petrykivd, line 22: You need to follow the same style code everywhere, so if you want brackets on separate lines, it should be like this at the beginning and at the end

petrykivd (approved):

PR #1828, opened 2024-11-25
Solution

bogdAAAn1:


app/main.py
```diff
@@ -1,13 +1,17 @@
...
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
```
KseniiaLog, line 10: You can use different names for variables to avoid repeating the "error" in functions. It's not critical, but maybe then there will be no confusion. I'm accepting the assignment, just writing a comment for the future.

KseniiaLog (approved):
You can use different names for variables to avoid repeating the "error" in functions. It's not critical, but maybe then there will be no confusion. I'm accepting the assignment, just writing a comment for the future.

PR #1852, opened 2024-11-27
Task completed

Arty-tech-dotcom:


app/main.py
```diff
@@ -1,13 +1,24 @@
...
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8",
+    }
     pass
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
+    status = "failed" if errors else "passed"
```
Dogober: All functions must contain only the return keyword.

    return {
        "errors": [
            format_linter_error(error) for error in errors
        ],
        "path": file_path,
    }

app/main.py
```diff
@@ -1,13 +1,24 @@
...
+        "path": file_path,
+        "errors": [format_linter_error(error) for error in errors],
+        "status": status,
+    }
     pass
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
+    return [format_single_linter_file(file_path, errors) for file_path, errors in linter_report.items()]
```
Dogober: for file_path, errors => for (file_path, errors)

Dogober (commented):
You need to add these changes

app/main.py
Professor-Douel: make your line shorter

Professor-Douel (changes requested):
Make line:13 shorter

PR #1886, opened 2024-12-06
Solution

Hlibynka:


app/main.py
```diff
@@ -1,13 +1,45 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
```
pavlopro: you should use `format_linter_error `method here

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+        ],
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
```
pavlopro: you should use `format_single_linter_file `method here

pavlopro (changes requested):

PR #1915, opened 2024-12-20
implement format_linter_error, format_single_linter_file, format_lint…

yarzhuravlov:
…er_report functions

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -14,7 +14,7 @@ def format_single_linter_file(file_path: str, errors: list) -> dict:
             format_linter_error(error) for error in errors
         ],
         "path": file_path,
-        "status": "passed" if len(errors) == 0 else "failed"
+        "status": "failed" if errors else "passed"
     }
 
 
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,25 @@
...
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error) for error in errors
+        ],
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed"
```
oliasymak24: Just `if errors else "failed"` will be enough

oliasymak24 (changes requested):
Small comment :)

app/main.py
yarzhuravlov: made changes, thanks for the advice

oliasymak24 (approved):

PR #1930, opened 2024-12-31
Feat: Add linter configuration

kram3ko:
Fix: Apply linter rules and format code
Solution: Implement task with linter and formatter

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -33,7 +33,9 @@ def format_single_linter_file(file_path: str, errors: list) -> dict:
 def format_linter_report(linter_report: dict) -> list:
     #  """
     # formats all errors for all report files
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1,13 +1,41 @@
...
-    # write your code here
-    pass
+    #  """
+    # formats all errors for all report files
+    # :param linter_report: Dict - result of worked
+    # func : format_single_linter_file
+    # :return: formated list with both funcs
+    # """
+    return [format_single_linter_file(key, value)
+            for key, value in linter_report.items()]
```
oliasymak24, line 41: Try to use more informative names for variables instead of key, value)

oliasymak24 (approved):
Good job)

PR #2057, opened 2025-02-12
'Solution'

Roksolana-K:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed" if errors != [] else "passed"
```
PrimeGlorious, line 15: Here's a small suggestion: instead of checking if errors != [], you could simply write if errors. This would make the code a bit cleaner and more Pythonic since an empty list evaluates to False.

PrimeGlorious (approved):
Great job on the code! However, I would suggest using a more descriptive commit message instead of "Solution".

app/main.py
Roksolana-K: Ok, thanks for the recommendation! You are right.

PR #2125, opened 2025-03-04
solution

Tolik-code:


app/main.py
```diff
@@ -1,13 +1,26 @@
...
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors":
+            [
+                format_linter_error(error) for error in errors
+            ],
+        "path": file_path,
+        "status": "failed" if len(errors) else "passed"
```
MNDonut: You can use just `if errors` because for python list returns `True` if the list isn't empty

MNDonut (changes requested):

MNDonut (approved):

PR #2322, opened 2025-05-30
implemented task

GrishaSeniv:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if len(errors) == 0 else "failed"
```
MNDonut: or you can reverse it:
`failed" if errors else "passed`
If a list,dict,set has at least one object, python considers it as `True`. So in this case if there are some errors, `if errors` will return `True`

MNDonut (changes requested):

Alkaponees (approved):
LGTM!

PR #2346, opened 2025-06-12
Solution

DrGregoryHouse4676:
Check my pull

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,13 +1,24 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
+        "column": error["column_number"],
+        "message": error["text"],
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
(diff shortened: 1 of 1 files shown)


DrGregoryHouse4676:
check again please

app/main.py
```diff
@@ -9,11 +9,10 @@ def format_linter_error(error: dict) -> dict:
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    format_errors = [format_linter_error(error) for error in errors]
     return {
-        "errors":format_errors,
+        "errors": [format_linter_error(error) for error in errors],
         "path": file_path,
-        "status": "failed" if errors else "passed"
+        "status": "passed" if not errors else "failed"
```
zqpiq, line 15: You don't need to use the if not construction, it's enough to use just if, because otherwise, if there are errors, you will still get a 'passed' status.

zqpiq:
Nice

zqpiq (approved):

PR #2347, opened 2025-06-12
Solution

zqpiq:


app/main.py
```diff
@@ -1,13 +1,32 @@
...
-    pass
+    return [
+        {"errors": [{"line": error["line_number"],
+                     "column": error["column_number"],
+                     "message": error["text"],
+                     "name": error["code"],
+                     "source": "flake8"}
+         for error in linter_report[key]]
+         if linter_report[key] else [],
+         "path": key,
```
MNDonut: 1. At the bottom of each task there is a checklist and in this task there is this rule:
image
2. Let's use another way to format the code. The general rule is: there are no characters after the opening bracket, so we move all code on the next lines and the closing bracket should be on a separate line without any characters before it. So for the first function it should be:

```
{
    "line": error["line_number"],
    "message": error["text"],
    "name": error["code"],
    "source": "flake8"
}
```

MNDonut (changes requested):

MNDonut (approved):

PR #2362, opened 2025-06-16
Solution

Orixara:


Otst0y (commented):
Looks ok to me. Good job, keep it up.

app/main.py
```diff
@@ -1,13 +1,23 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        **{new_key: error[old_key] for old_key, new_key in {
```
yagginov: Everything looks fine except for this. I think it's an unnecessary complication of the code that reduces readability.

And also this format of brackets is called a bad example in the file. 
link: [Сheck Your Code Against the Following Points](https://github.com/yagginov/py-linter-formatter/blob/master/checklist.md)

yagginov (commented):
Everything looks fine, but these lines I marked are very strange. I think it's better to do a direct mapping there. like:
```
        "line": error["line_number"],
        "message": error["text"]
```

yagginov (changes requested):
I think this is still worth changing.

app/main.py
Orixara: Hi! Thank you, I think about it too, but I thought that everything had to be with comprehensions, that's why I did this strange code😅

yagginov (approved):
All good.

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [format_linter_error(error) for error in errors],
+            "path": file_path,
+            "status": "passed" if not errors else "failed"}
```
MNDonut: let's follow the same code styling as in other functions, so that brackets are on separate lines
also you can simplify if statement using direct condition "if errors"

MNDonut (changes requested):

app/main.py
Orixara: Thank you!

MNDonut (approved):

PR #2373, opened 2025-06-17
Solution

mishahoshii:


app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "passed" if not errors else "failed"
```
Orixara: It can be simpler to read if we use only "failed" if error else ...

app/main.py
```diff
@@ -1,13 +1,21 @@
...
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(file_path, errors)
```
Orixara: There it would be better to reformat code:
return [
    ....
    ....
]

Orixara (changes requested):
It's great job, but you have some little things that you can improve

app/main.py
mishahoshii: Thanks for your advice

app/main.py
mishahoshii: Ok, thank you

mishahoshii:
Resolved

mishahoshii:
Resolved

Orixara (approved):

Orixara (approved):

PR #2374, opened 2025-06-17
Solution

MakovM:


app/main.py
```diff
@@ -1,13 +1,30 @@
...
-    # write your code here
-    pass
+    return {
+        "errors": [{
+            "line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"],
+            "name": error["code"],
+            "source": "flake8"
+        }
```
yagginov: You should have reused the code here. You already have the format_linter_error() function, that's what you need to use here, not copy the logic again.

yagginov (changes requested):
Everything else seems to be correct, both in code and formatting. Just fix the place I showed.

yagginov (approved):
All good

yagginov (approved):

PR #2570, opened 2025-09-17
sol

delirium95:


app/main.py
```diff
@@ -1,13 +1,23 @@
...
+        "path": file_path,
+        "status": "failed" if errors else "passed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [format_single_linter_file(path, errors)
+            for path, errors in linter_report.items()]
```
petrykivd: Good work! Your code is correct, but there are two small style issues:
1. Inconsistent formatting

Functions 1 and 2: return is on a separate line
Function 3: return is on the same line

Pick one style and use it everywhere. I suggest the separat-line style ;)
2. First function is a bit overcomplicated
The dictionary comprehension in format_linter_error works, but it's hard to read. It's not critical, but simpler code is better code.

Remember: Keep your code style consistent! 👍

petrykivd (changes requested):
Good job! 👍 

Code style is very important, so let's make it a habit to always follow it in all parts of the code.

oliasymak24 (approved):

PR #2662, opened 2025-10-17
implemented format_linter_error(), format_single_linter_file(), forma…

vakt159:
…t_linter_report()


VadymVynokur (approved):
in the first function it is not necessary to do this loop,
u can just
return {"line": error["line_number"],
            "column":.....}
  and so on
  in this case it will not be necessary to also use "|"
  
  BUT
  if it works - it works, so gret job!

PR #2675, opened 2025-10-21
Solution

dmytro-malyk-dm:


app/main.py
```diff
@@ -1,13 +1,24 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+
```
ohorodnykostap, line 2: If you use an empty line before the return, you should use it in other functions as well or not use it at all.

ohorodnykostap (approved):
looks good!

PR #2688, opened 2025-10-22
Solution

Akcolotle:


Guillelmina8 (approved):
Good job! Looks better than I could imagine.

PR #2691, opened 2025-10-22
'solution'

DarynaN:


app/main.py
```diff
@@ -1,13 +1,31 @@
...
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [format_linter_error(error) for error in errors],
+        "path": file_path,
+        "status": "failed"
+        if [format_linter_error(error) for error in errors]
```
vadym-zahrebelnyi: You could simplify this by just checking the length of `errors` instead of iterating over them again — no need to call `format_linter_error` a second time here.

vadym-zahrebelnyi (approved):
Good job!

app/main.py
DarynaN: Thanks for your advice!

PR #2692, opened 2025-10-22
Solution

tigersaymon:


app/main.py
```diff
@@ -1,13 +1,43 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {
+        "line": error["line_number"],
```
Iventyk: Better use the '.get() ' method for all values cases when there missing value

app/main.py
```diff
@@ -1,13 +1,43 @@
...
+            "status": "failed" if errors else "passed"
+            }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [
+        {
+            "errors": [{
```
Iventyk: Different code style: compare lines 12-13 and 28-30 and choose the better style

Iventyk (changes requested):
Hi! I noticed that the code style is inconsistent across the file.
There are different formatting patterns, such as parentheses on different lines.
Please make the code style uniform and consistent according to one standard, so it’s easier to read and maintain.
Thank you!

app/main.py
tigersaymon: Done. Replaced direct dictionary access with .get() for all error fields to prevent KeyError on incomplete reports

app/main.py
tigersaymon: Thanks for the comment! I consolidated all functions to a single indentation style to ensure full code consistency

DarynaN (approved):

PR #2897, opened 2026-01-06
Solution by Kryvutsa

pawunder1992:


app/main.py
```diff
@@ -1,13 +1,26 @@
...
+            "column": error.get("column_number"),
+            "message": error.get("text"),
+            "name": error.get("code"),
+            "source": "flake8"}
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {"errors": [{"line": mis.get("line_number"),
```
SerhiiHryniuk: Better call function format_linter_error than write again this all block of code

app/main.py
```diff
@@ -1,13 +1,26 @@
...
+                        "name": mis.get("code"),
+                        "source": "flake8"} for mis in errors],
+            "path": file_path,
+            "status": "failed" if len(errors) else "passed"}
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [{"errors": [{"line": mis.get("line_number"),
```
SerhiiHryniuk: Here it is also better to call function format_single_linter_file

SerhiiHryniuk (changes requested):

app/main.py
pawunder1992: Thanx for checking, resolved

app/main.py
pawunder1992: Thanx for checking, resolved

SerhiiHryniuk (approved):

PR #2924, opened 2026-01-14
Solution

AndrewGalynin:


AndrewGalynin:
luke is still checking my homework on website, but approved it on github

PR #2939, opened 2026-01-17
soluiton for kodree

gunsts:


gunsts (commented):
changed according to your review suggestion.

vadim-ilchenko (approved):

PR #3015, opened 2026-02-14
Solution

Reverad:


app/main.py
```diff
@@ -1,13 +1,26 @@
...
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [
+            format_linter_error(error)
+            for error in errors
+        ],
+        "path": file_path,
+        "status": "failed" if len(errors) > 0 else "passed"
```
pawunder1992, line 18: You can write ''if errors'' instead of "if len(errors)" but it is not a big problem, good solution.

pawunder1992 (approved):

PR #3133, opened 2026-03-20
Solution

lubkoserhii:


alisahoys (approved):
Good use of list comprehensions ([format_linter_error(error) for error in errors])
Using "failed" if errors else "passed" is a clean Pythonic way for conditional status.

PR #3158, opened 2026-03-25
Solution

Oleksandr002:


ratatuiii (approved):
Everything done correct, within "return statement only" rule

PR #3173, opened 2026-03-26
dd

olha12232:


andriihrechko (approved):

dimaaa-cmd (approved):
That's all correct, except for one thing: the requirements.txt file shouldn’t be changed. Since you're using a virtual environment, the interpreter version and dependencies are expected to remain consistent. You're working on a specific task, and in real-world scenarios, you'll typically need to stick with the chosen Python version - especially if the project already has existing functionality.

andriihrechko (approved):

PR #3446, opened 2026-07-17
Success

andriynavr07:


Sn1kls (approved):
Approved. One small optional note:
`errors: list` could be typed more precisely as `list[dict]`

PR #3488, opened 2026-08-03
Solution

MaksimDonchyk:


app/main.py
```diff
@@ -1,13 +1,16 @@
 def format_linter_error(error: dict) -> dict:
-    # write your code here
-    pass
+    return {"line": error["line_number"],
+            "column": error["column_number"],
+            "message": error["text"],
```
vlad-krivospitsky: Typically, the opening bracket is placed on the first line, and the content is indented by four spaces. The closing bracket is placed on a separate line

For example:

    return {
        "line": error["line_number"],
        "message": error["text"],
        "name": error["code"],
        "source": "flake8"
    }

MaksimDonchyk: fixed

vlad-krivospitsky (approved):

PR #3489, opened 2026-08-03
Solution

CemXokenc:


app/main.py
```diff
@@ -1,13 +1,34 @@
...
+        "name": error["code"],
+        "source": "flake8"
+    }
 
 
 def format_single_linter_file(file_path: str, errors: list) -> dict:
-    # write your code here
-    pass
+    return {
+        "errors": [{
```
Avalanche-one: Duplicated lines or whole blocks of code for creating formated dict. You can use your previous function to avoid it.

app/main.py
```diff
@@ -1,13 +1,34 @@
...
+        "path": file_path,
+        "status": "passed" if not errors else "failed"
+    }
 
 
 def format_linter_report(linter_report: dict) -> list:
-    # write your code here
-    pass
+    return [{
+        "errors": [{
```
Avalanche-one: The same here - duplicated lines of code

Avalanche-one (changes requested):
Look into the checklist file, which is at the end of README file. Make sure that your solution fits well it's requirements

app/main.py
CemXokenc: Fixed. But I think my variant was more illustrative

app/main.py
CemXokenc: Same

Avalanche-one (approved):

